PR #13, opened 2015-01-17
Add DNS request

lukasz-madon:


README.rst
```diff
@@ -50,7 +50,8 @@ DNS lookup...
   the lookup.
 * If ``gethostbyname`` does not have it cached then a request is made to the
   known DNS server that was given to the network stack. This is typically the
-  local router or the ISP's caching DNS server
+  local router or the ISP's caching DNS server.
+* Port 53 is opened to send UDP request to DNS server.
```
alex: Should this maybe mention that if the request might become a TCP request if it's too large / other details?

lukasz-madon: If I understand correctly, DNS requests can become TCP only in case of request like zone transfers which are server-server (replication). It's very unlikely for browser to cause that.

alex: From wikipedia: `The Transmission Control Protocol (TCP) is used when the response data size exceeds 512 bytes, or for tasks such as zone transfers. Some resolver implementations use TCP for all queries.`

I think we should at least mention something like `(if the response size is too large, TCP will be used instead).`

lukasz-madon: I check http://serverfault.com/questions/181956/is-it-true-that-a-nameserver-have-to-answer-queries-over-tcp and RFC and indeed browser request may use TCP.

PR #21, opened 2015-01-17
Update README.rst

tig:
Explained what happens from the HID->Windows app layer.


alex:
Awesome work, thanks!

Issue #50, opened 2015-01-17
Repo Refactoring


Trozz:
I would suggest that we look at creating individual files for each question, this will allow easier management of contents.

EaterOA:
I support the notion that the repo reeeeeeally needs some refactoring.

Whether this means better organization in one file or creating individual files, I don't know. At the very least, it needs some clearly defined sections. For instance, it can be made:

* Time-based, e.g. before first packet leaves host, after first packet leaves host, after first packet arrives at destination, etc.
* Protocol-based roughly following the order in which they're invoked, e.g. DNS, UDP, IP, ARP, etc.

phildini:
See #51, also:

My $0.02 is that time-based is how most people organize this information in their heads, and best matches the "answering the interview question" approach. That said, if some protocol happens at multiple steps, links should be our friend here.

kojiromike:
I think that layer-based is going to be the most beneficial for review. As this documentation gets more and more comprehensive, some people (like me) are going to want to gloss over, say, the physical layer, and focus more on transport->application. Others are more interested in the EE aspects and will want a lower-level point of view. They may not care what HTTP has to say about it.

I'm not saying it should try to adhere strictly to OSI or anything like that, but _trending toward_ a layer based breakdown has obvious benefits for someone trying to gain a deep understanding of, say, the application aspects of the interaction, without caring about voltage drops.

PR #51, opened 2015-01-17
restructuring #50

Trozz:
Moved text to separate files.


alex:
There's a merge conflict here, can you please merge/rebase this.

phildini:
In general, I think I'm +1 on the idea of splitting the sections into multiple files, so long as we can:

- Keep the correct ordering
- Have "previous step" and "next step" links in the section files. 

I'm pretty sure that's not feasible through github; perhaps this should be a ReadTheDocs thing? On the one hand, I like one link that people can visit to see all the content, and github makes it semi-clear how to participate.

On the other hand, pretty formatting, and potentially easier navigation.

joaohencke:
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

PR #57, opened 2015-01-17
Update README.rst with basic HTTP

mossberg:
Nothing fancy, just basic HTTP stuff.


driusan:
I'm terribly biased, but I prefer my pull request #58 which goes into a little more detail.

mossberg:
I prefer yours too, excellent work :)

PR #78, opened 2015-01-17
Add more detail to page rendering and add links to specs.

samlh:


README.rst
```diff
@@ -55,7 +55,7 @@ converts the HID usage into a scancode. In this case the scan code is
 ``VK_RETURN`` (``0x0D``). The ``KBDHID.sys`` driver interfaces with the
 ``KBDCLASS.sys`` (keyboard class driver). This driver is responsible for
 handling all keyboard and keypad input in a secure manner. It then calls into
-``Win32K.sys`` (after potentially passing the message through 3rd party
```
dtracers: I do believe this is a revert.

samlh: Fixed in https://github.com/samlh/what-happens-when/commit/707ec21e1e3742bc74686782be979aeb9026127e

samlh:
The travis failure is for a term from the css spec, so cannot be addressed.

alex:
That's pretty obnoxious! Can you just remove the `--weasel`` flag for now.

PR #82, opened 2015-01-18
Expand USB interface sections.

tdaede:
I wasn't really happy with the description of the physical USB interface, so I rewrote most of it. This could probably use more editing still, but I feel that is better than what was there before.


phildini:
Would you mind looking at the newly-expanded USB section and seeing what you think? It could probably still use edits, but seems to mostly cover what you have here.

Thanks!

PR #85, opened 2015-01-18
include info on layer 1-3 of the osi model

thisissmithj:


README.rst
```diff
@@ -131,39 +131,35 @@ DNS lookup...
...
+both sides of the connection.
 
-This send and receive happens multiple times following the TCP connection flow:
+TCP packets
+~~~~~~~~~~~
 
 * Client chooses an initial sequence number (ISN) and sends the packet to the
   server with the SYN bit set to indicate it is setting the ISN
-* Server receives SYN and if it's in an agreeable mood:
+* Server receives SYN and if a service is listening on that port:
```
ikanc: Add a note on firewall allowing the connection on that port? :-)

phildini:
These changes look good, but HEAD has moved on without you. Would you mind fixing the merge conflict and trying again?

thisissmithj:
Sure, I'll update the PR. I'll also include a brief note on firewalls as phildini suggests.

thisissmithj:
Updated!

phildini:
I'm having some hesitation merging this, because while it looks like you're adding a bunch of really useful OSI Layer info, it looks like there's some information in what you're removing that isn't being captured by your additions. (Specifically, what is added to the packet at each layer by the OS)

Am I mis-reading? Thanks for your patience, and your contributions!

PR #92, opened 2015-01-18
Add location of hosts file on Windows

JoeMalt:


phildini:
Would you mind adding the locations for other OSes? Also, tests seem to be failing.

Kwpolska:
@phildini  Linux/Unix is `/etc/hosts`, that’s already handled by the previous word on the line.   OS X, iOS and Android have `/etc/hosts` as a symlink to a fancy system directory, so saying `/etc/hosts` is good enough.  And let’s assume other OSes don’t exist.

@JoeMalt The line is too long, you must break it into 79 characters or less.  Like this:

```rst
  ``/etc/hosts`` file (``C:\Windows\system32\drivers\etc\hosts`` on Windows),
  before trying to resolve the hostname through DNS.
```

mburns (approved):

PR #116, opened 2015-01-20
Various improvements: typos, spelling and grammar

frankgeerlings:
There's actually room for some more improvements, but these are the ones that ought not to cause any form of discussion.


phildini:
Thanks for the fixes!

PR #121, opened 2015-01-20
Typo

iamDeveloper:
denounces instead of debounces


phildini:
I'm fairly certain debounces is the correct terminology.

aprescott:
I don't think this correction is right. https://en.wikipedia.org/wiki/Switch#Contact_bounce

iamDeveloper:
Oh yeah, Right.

PR #123, opened 2015-01-21
Update DAC to modem

fehwalker:
Fixed up the "where it leaves the computer" bit to use modem details instead of DAC.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -303,10 +303,10 @@ your computer, possibly through a local network, and then through a modem
 signal suitable for transmission over telephone, cable, or wireless telephony
 connections. On the other end of the connection is another modem which converts
 the analog signal back into digital data to be processed by the next `network
-node`_ where its from and to addresses would be analyzed further.
+node`_ where the from and to addresses would be analyzed further.
 
-Most larger businesses and some newer residential connections will have fibre
-or direct Ethernet connections, in which case the data remains digital, and
+Most larger businesses and some newer residential connections will have fiber
+or direct Ethernet connections in which case the data remains digital and
 is passed directly to the next `network node`_ for processing.
 
 Eventually, the packet will reach the router managing the local subnet. From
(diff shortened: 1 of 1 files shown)


Thugg55:
When you press the g-key, it depends on the context in which it is pressed. If you are on a webpage or in a search bar, pressing the g-key might initiate a search using Google as the default search engine

DjibbeDiguina:
I've submitted a pull request to the "what-happens-when" repository to contribute a comprehensive answer to the question "What happens when you type in that domain name?" My contribution aims to enrich the existing content with detailed insights into the DNS resolution process, TCP/IP communication, firewall interaction, HTTPS/SSL encryption, load-balancing mechanisms, web server functionality, application server involvement, and database interactions.

By providing a thorough explanation of each step involved in serving web content, I aim to enhance the understanding of readers and help them grasp the intricacies of the web stack. This contribution aligns with the repository's goal of compiling comprehensive answers to common questions about internet technologies.

DaBetterman:
What happens when the g key is pressed?

Depending on where you find yourself on your pc...
if your pc just started and you are on linux, pressing g will search for an icons, files, etc on your desktop starting with a 'g'. Google chrome it will basically do the same as to look for a recent bookmark that starts with g.
If I have to explain it tehnically then basically what happens is that a computer generally just understands '1' and '0'. According to the ASCI chart the binary value of 01100111 represents the symbol or character 'g'. So when you press the key 'g' and electronic signal is sent to the OS which will display the binary code to the symbol 'g' and that explains it for each key on the keyboard.

2012Inga:
when pressing the g key it depends, In summary, the action triggered by pressing the "g" key depends entirely on the context in which it's pressed, whether it's a text entry, a shortcut, a gaming command, or a system-wide action.

Akpos-Ben:
In essence, pressing the "G" key sets off a sequence that culminates in a response determined by the operating system and any active applications.

PR #136, opened 2015-01-27
explanation about mandatory connection header

dhaowoods:
HTTP/1.1 applications that do not support persistent connections MUST include the "close" connection option in every message.


BenjaminProgram:
Please fix the trailing whitespace on line 421 and the tabulation on line 405.

BenjaminProgram:
I fixed the build errors in pull request #148

PR #139, opened 2015-01-28
Make gender neutral.

mewmew:
Use *their* instead of *he or she* as individuals may define their gender in a diverse variety of ways.


alex:
Thanks!

PR #142, opened 2015-01-28
HTML parsing information updated

logeshpaul:
Updating a few information about HTML parsing, Will write more in detail. 

+1 for creating the repo.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -481,7 +481,11 @@ HTML parsing
 
 The primary job of HTML parser to parse the HTML markup into a parse tree.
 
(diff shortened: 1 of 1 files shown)


phildini:
These updates look great! 

One fix: You remove the lines about fetching external resources and executing JS. Could you add those back to your changes in some way, so that there is no removal of information?

(Those topics are probably covered by the HTML5 spec, but it helps to have mentions in this document)

logeshpaul:
@phildini I thought of creating a separate section for fetching content and resources. As of now I have added those content back. 

In next PR I will add more details about CSS parsing, Rendering and Reflow.

Thanks!

phildini:
If you're going to fill out those sections more, mega +1 from me. I'm just aiming to never lose detail with any PR.

Thanks!

PR #157, opened 2015-02-18
Fix typo in readme

YusefOuda:
Change 'effecting' to 'affecting' inside of Post-rendering section


phildini:
I think this is actually the correct usage. The 'effect' is another round of rendering, so that's what's being 'effected'. We are not 'affecting' the rendering.

YusefOuda:
I agree, I knew effecting didn't sound correct but wasn't sure of the best word to replace it. I agree with causing

PR #163, opened 2015-03-12
add some acks

revmischa:
cheers!


revmischa:
no love?

GortMcTabs:
Although I'm a fan of the buzzword replacer, it's not going to get any love when it comes to a PR.

revmischa:
What about just the original work?

PR #175, opened 2015-04-27
Fixed a typo

richardartoul:
I removed the letter "s" from the phrase "uses a custom parsers" so that it now reads "uses a custom parser"


kamalkabirsaini:
Yes it will definitely work on

PR #178, opened 2015-05-21
Changed dropbox to dropdown box

manish-shrivastava:
Changed dropbox to dropdown box, because as per my understanding it is a div dropdown and not dropbox.


mburns (approved):

mburns:
this was fixed in #285

PR #180, opened 2015-06-06
Add linux boot process

rhoml:


README.rst
```diff
@@ -676,3 +676,33 @@ page rendering and painting.
...
+-------------------
+
+There are 6 elements involved on the Boot process of a Linux box:
+
+1. **Bios**
+  * Perform system integrity checks.
+  * Search, Load and execute the Boot loader program.
+  * Scans for devices (Hard Drives, cd-rom, Floppy disk) once it detects the boot loader it loads it to memory.
+
+2. **MBR**
```
daodennis: This is a bit outdated. GPT is a more modern partioning scheme.

rhoml: Actually yes, but wanted to do the more generic one. Would it make sense to add both or just GPT as the latest?

PR #182, opened 2015-08-25
AS definition

lcafaro:


matkoniecz:
@alex Can you consider either merging it or rejecting it?

PR #200, opened 2016-05-17
Bring up the parallel JS loading limitations

alexpanov:


README.rst
```diff
@@ -585,7 +588,9 @@ The algorithm consists of two stages: tokenization and tree construction.
 **Actions when the parsing is finished**
 
 The browser begins fetching external resources linked to the page (CSS, images,
-JavaScript files, etc.).
+JavaScript files, etc.). However, while the browser loads a script any content 
+introduces on in the HTML file below the script is blocked until the download is 
+complete. This is true even for the downloads from other domains.
```
frewsxcv: > any content introduces on in the HTML file below the script is blocked until the download is complete

This isn't true if the `async` attribute is set on `<script>`, right?

For reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script

PR #202, opened 2016-05-20
Added clarfiying line break details to HTTP header

braeden123:
https://www.w3.org/Protocols/rfc2616/rfc2616-sec2.html#sec2.2
"\r\n, because it's defined as the line break in the protocol specification."


braeden123:
Build failure is not the result of this commit.

PR #208, opened 2016-05-21
Put the parse URL section in correct order

fhightower:
Switched the order of the "parse url" and "Is it a URL or a search term?" sections to reflect how the process would occur chronologically.


frewsxcv:
In order to find out if the text in the address bar is a 'search term or URL', URL parsing needs to happen. I imagine if URL parsing fails, the address bar text will be treated as a search term.

I think the wording in these sections could better reflect that, but I don't think it's necessary for them to be switched right now.

PR #226, opened 2016-11-28
Update for HTTP/2

hadibrais:
SPDY has been deprecated and replaced by HTTP/2. Most browsers now will be using HTTP/2 to get the Google web page. The changes in this commit explain what happens when HTTP/2 is used by the client and the sever to communicate.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -368,13 +368,13 @@ This send and receive happens multiple times following the TCP connection flow:
 TLS handshake
 -------------
 * The client computer sends a ``ClientHello`` message to the server with its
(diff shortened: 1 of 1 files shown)


coderobe:
Actually, after the initial request, opera (v16 and up), chrome/chromium, & derivates and potentially a few other browsers will attempt to connect using QUIC instead of HTTP/2 if the server advertises it being available

hadibrais:
Oh, I am not aware of QUIC. I am using the latest version of Firefox which uses HTTP/2. So I guess we need to consider both protocols depending on which version of which browser is being used.

Issue #225, opened 2016-11-28
UDP port 53


marauderlabs:
In section **ARP process** , 
~~Port 53 is opened~~ to send a UDP request to DNS server (if the response size is too large, TCP will be used instead).
The above sentence is phrased wrong and is misleading to signify that Port 53 is opened to send a UDP *request*.
UDP 53 is the destination port. It's already open in the DNS server. A request _to_ the port is initiated.

PR #235, opened 2017-03-22
Set correct client ports for the DNS UDP query + Cache Poisoning

drakelevy:
#230 is already aimed at fixing this but is poorly worded and vague.

I included the range of ports that the client could open and if they're on Windows what the specific range was. I also included a reference to DNS cache poisoning so there's some context on why the client is randomizing their port.

This also addresses issue #225


Amelianancy:
There is conflict

PR #244, opened 2017-07-30
expanding when 'g' is pressed

orsenthil:
* dropdown instead of dropbox.
* program is run on thousands of computers (it's a distributed system).


thesohelshaikh (approved):
Fixed typos in when ‘g’ is pressed section.

PR #256, opened 2017-09-17
Add code of conduct

jamiehannaford:
This is a super popular repo so it'd be great if we could have a CoC for folks to consult and rely on where necessary. Taken from the [Contributor Covenant](https://www.contributor-covenant.org/version/1/4/code-of-conduct/).

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,73 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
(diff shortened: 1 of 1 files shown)


jamiehannaford:
@alex would it be possible to get this merged? 🙇  If not I can close.

HybridDog:
l've just found out that you can put emoticons into source code in github comments:
```C
while (i <= 9) {
	printf("this is 🙇 the thing\n");
}
```

PR #259, opened 2017-10-03
Update README.rst

thehackercat:
Add load balancer processing after getting Http requests.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -484,9 +484,9 @@ and IIS for Windows.
   URL Rewrite for IIS), it tries to match the request against one of the
   configured rules. If a matching rule is found, the server uses that rule to
   rewrite the request.
-* If the server is configured with a load balancer, the request is forwarded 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -484,6 +484,10 @@ and IIS for Windows.
   URL Rewrite for IIS), it tries to match the request against one of the
   configured rules. If a matching rule is found, the server uses that rule to
   rewrite the request.
+* If the server is configured with a load balancer, the request is forwarded
+  according to the configured balance rule. The reverse proxy load balancing
+  is the most common in the request of the proxy server to forward the request
+  to the upstream and downstream back-end server for reducing server pressure.
```
frewsxcv: > The reverse proxy load balancing is the most common in the request of the proxy server to forward the request to the upstream and downstream back-end server for reducing server pressure.

This feels wordy to me. Wonder if we can break it into a couple sentences and make it more approachable. 🤔

PR #301, opened 2020-04-13
To add info about how to customize the autocompletion function

gogomillan:
The autocompletion function it is a current relevent function in the modern browsers as an expression of Artificial Intelligence. So can tune it is important!

Surces: 
https://wiki.mozilla.org/Firefox/URL_Bar_Algorithm
https://support.mozilla.org/en-US/kb/address-bar-autocomplete-firefox?redirectlocale=en-US&redirectslug=Location+bar+autocomplete#w_changing-results-on-the-fly


gogomillan:
Taking into account the AI in the workflow.

PR #270, opened 2018-07-31
fix: replace IP with MAC in ARP process

AjeetK:
In ARP process, it tries to find out MAC address of the interface.

PR #183, opened 2015-09-12
Typo? Should be MAC address. ARP is the discovers the MAC address

hhsue-zz:

PR #304, opened 2020-08-23
ARP resolves an IP address to a MAC address

j0le:
not the other way around. (one word changed)

The IP addresses of default gateway and DNS sever are already known before the "ARP process", because they were configured manually or via DHCP or in case of IPv6 for example via router advertisement.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -287,7 +287,7 @@ Switch:
     Target MAC: interface:mac:address:here
     Target IP: interface.ip.goes.here
 
-Now that the network library has the IP address of either our DNS server or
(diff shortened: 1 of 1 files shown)


j0le:
dupplicate of #183 and #270

technicalgyandeep:
What is it

j0le:
@technicalgyandeep
What is what? What exactly do you mean?

yezz123 (approved):

mallika-b:
Ugh, why has this not been merged? 
I found it utterly confusing.

akostadinov (approved):

PR #306, opened 2020-08-26
Updated the URL parse section

Cirine-S:
I better explained the different parts of URL, which mainly are: protocol, subdomain, domain name and the path to a file.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,7 @@
 What happens when...
 ====================
 
-This repository is an attempt to answer the age old interview question "What
+This repository is an attempt to answer the age old interview questiond "What
 happens when you type google.com into your browser's address box and press
 enter?"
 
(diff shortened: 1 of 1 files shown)


emmaudeji:
When you enter a website address (URL) on your browser (e.g Chrome), the browser uses this to search for the I.P address of the website from another location on the internet called DNS (Domain Name System). When it succeeds in retrieving the I.P address, it again searches the internet using this address for a server location where the content of your request was stored by the website host. When it retrieves it through a channel called Transmission Control Protocol/Internet Protocol (TCP/IP) suites, it then displays the content on your device. You will see these as web pages, but behind the hook, these are text files formatted in a way that your browser can interpret them. This format is known as HTML.

what_happens_when_you_enter_google com_in_your_browser__alx_holberton
.

muhammadthani:
When you type https://www.google.com/ in your browser and press Enter, a DNS request is sent to a DNS server to translate the domain name "[www.google.com](http://www.google.com/)" into an IP address.

After the IP address is obtained, a TCP/IP connection is established between your computer and Google's web server.

The connection passes through any firewalls that may be in place and is secured using HTTPS/SSL encryption to protect any data that is transmitted between your computer and the server.

The request is then sent to a load-balancer, which distributes the request to one of several web servers in Google's network.

The web server retrieves the requested web page and any necessary data from an application server and/or database, and sends the page back to your browser, where it is displayed on your screen.

muhammadthani:
When you type https://www.google.com/ in your browser and press Enter, a DNS request is sent to a DNS server to translate the domain name "[www.google.com](http://www.google.com/)" into an IP address.

After the IP address is obtained, a TCP/IP connection is established between your computer and Google's web server.

The connection passes through any firewalls that may be in place and is secured using HTTPS/SSL encryption to protect any data that is transmitted between your computer and the server.

The request is then sent to a load-balancer, which distributes the request to one of several web servers in Google's network.

The web server retrieves the requested web page and any necessary data from an application server and/or database, and sends the page back to your browser, where it is displayed on your screen.

PR #308, opened 2021-01-01
Add a section on TCP congestion control

jyn514:
I didn't go into too much details on the differences between algorithms or when congestion control performs badly (high throughput, high packet loss), but I can if you like :)


alex:
There's some long lines the linter is unhappy about.

Thanks!

alex (approved):

PR #360, opened 2022-05-22
added item on common browser UI

Lukong123:


andrew21-mch:
looks good Anne

Lukong123:
> looks good Anne

Thanks

PR #373, opened 2022-10-08
What happens when you type google.com in your browser and press Enter

VikitorChidi:


Mzee1991:
What happens when you type google.com in your browser and press Enter

abrish-software:
What happens when you type google.com in your browser and press Enter
he average computer user doesn’t have to worry about what goes on behind the scenes when they access their favorite website, thanks to abstraction. In this series we’ll look at the processes that take place from the time you type your favorite website address on the browser and hit enter to the time the webpage is displayed as illustrated below.


Flow of the request created when you type https://www.google.com/ and press Enter
DNS Request
Every computing device on a network has a unique IP address that identifies it on the network. For IPv4 addresses, this is a set of dot separated values ranging from 0.0.0.0 to 255.255.255.255. While this is great for computers, humans cannot recall them very easily. For this reason, the Domain Name System (DNS) allows names such as google.com to be translated to IPs.

To make a connection, the domain name ‘www.google.com’ has to be resolved to an IP address. This happens in a series of steps as follows:

First, the browser checks its local DNS cache if it had recently retrieved it. If this fails, it checks the OS cache for the IP. If both don’t have the record, a request is made to the recursive DNS server, ussually located at the ISP.
Recursive DNS servers asks the root Name Servers which direct them to the TLD Name servers (such as .com in this case). The TLDs direct the queries to the respective Authoritative name servers that house the DNS records. The result is returned to the Recursive DNS server which caches it and forwards it to the client and finally the browser that also caches it.
Once the domain name has been resolved, the browser tries to establish a TCP connection with the respective server.

TCP/IP
The Transmission Control Protocol (TCP) is one of the most common, robust and reliable protocols for data transmission. Unlike UDP (User Datagram Protocol), TCP provides error and flow control by checking and resending lost packets. Your browser makes a TCP/IP connection to the resolved IP using the TCP/IP three-way handshake. This involves sending SYN (synchronize)and ACK (acknowledgement) packets between your browser and the server.

The client (Your browser/computer) sends a SYN packet asking it to make a connection.
If the server is available, it initiates a connection by sending an ACK packet back.
The client recieves and responds with an ACK packet thereby establishing a TCP connection.
Firewall
A firewall is a security implementation that monitors and filters network traffic based on laid out security policies. Once a connection is established with the server, a HTTP, or HTTPS in this case, request is made to the server requesting for the page matching the specified route in the URL (Uniform Resource Locator), in this case google.com’s homepage. A firewall is the first point of contact with corresponding server. Depending on the type of firewall, whether network or host-based, connections are either blocked or allowed based on policies. While network firewalls mitigate unauthorized access to private networks connected to the internet, host-based networks are installed on the host and configured for specified hosts. Essentially, firewalls regulate traffic to and from the target device on specified ports. In this case, https traffic on port 443.

HTTPS/SSL
Hyper Text Transfer Protocol Secure (HTTP) is a secure version of the HTTP protocol that uses the Secure Sockets Layer (SSL) encryption to encrypt all network traffic between a client and the server. Since the URL specified “https://” in the first part of the address, this protocol is used and all traffic is encrypted such that only the client and server can decrypt. An SSL certificate installed on the server, ussually the Load Balancer, ensures that https is used.

Load Balancer
Websites with lots of traffic such as google.com or [www.google.co](http://www.google.com/)m implement what is known as a load balancer to distribute requests to several other servers sitting behind the load balancer. Using a software such as HighAvailability Proxy (HAProxy) and an algorithm like round robin, client requests are distributed to the servers to avoid overload on a single server.

In our case, the https request made to [www.google.com](http://www.google.com/) is recieved by the load balancer listening on port 443. Using its algorithm, the request is sent to one of the available network of web servers that actually responds to the request.

Web Server
A web server is a software and/or hardware that uses HTTP or other protocols to respond to client requests. The main job of a web server is to display website content through storing, processing and delivering webpages to users. The main webservers in use as of this writing are Nginx and Apache. They serve static webpages and control how the rest of the URL (the part after the TLD ‘.com’) is interpreted and/or redirected. In case of a dynamic page, for instance a page that needs to fetch contents from a database or implements some businesss logic, the request is passed on to an Application server.

In our case, the load balacer passes the request to a webserver such as Nginx or Apache listening on port 80 for HTTP requests.

Application Server
An application server is a server specifically designed to run applications. It essentially provides the environment necessary to run the said applications and handles routes to access specified logic/application. These handle dynamic webpage requests, load-balancing, connections to the database, etc.

Database
A database is information that is set up for easy access, management and updating. Database servers provide access to the database, essentially the data stored through the application server. Database Management Systems such as MySQL, MongoDB. Postgres, etc. provide defined ways of accessing, managing, and updating the data.

With all these different components working together, webpages are retrieved and displayed on the client browser.

To recap, the browser makes a DNS request to resolve the domain name ‘www.google.com’ to an IP. It then makes a TCP/IP connection with the server (Load balancer), and a firewall monitors the traffic between the two. A load balancer forwards the request to one of the webservers which responds to our request. HTTPS is used to encrypt all traffic between our browser and the server. In case of dynamic content or logic, the request is sent to an application server which can connect to the database where necessary. The browser recieves the response and finally displays the [www.google.com](http://www.google.com/) webpage.

PR #374, opened 2022-10-08
my addition

zinkop:
This is my addition from the ALX project


mburns (changes requested):
content should go in `README.rst`

PR #381, opened 2022-10-09
Added more points on TCP section

Natiman58:
Added few pointes in the TCP connection setup section of README file


TaiwoForbes:
Added some point on TCP section

PR #383, opened 2022-10-09
Made some modifications to the DNS section

TaiwoForbes:


TaiwoForbes:
Some modifications to the DNS section

PR #403, opened 2022-10-09
Update README.rst

royalboe:
Added additional explanation to why the browser tries to check the hosts file on the OS while trying to resolve a URL


tomjes03:
Added additional explanation to why the browser tries to check the hosts file on the OS while trying to resolve a URL

PR #409, opened 2022-10-10
Load balancing

captjay98:
A brif explnation of what lod balancing is.

### DIFF
--- a/loadbalancer
+++ b/loadbalancer
@@ -1,0 +1,1 @@
+A load balancer as the name implies is used to balance load between servers.


twinnerhenock:
Lets try to add new documents regarding google.com

twinnerhenock (commented):
Disadvantages of load balancing: Although scalability is among the most appealing advantages of both cloud computing and load balancing, scalability is also among the most prominent weaknesses of load balancing. Scalability is hampered for most load balancers by a limited number of nodes for distributing processes. Other challenges include energy consumption, performance monitoring, QoS management, resource scheduling, and service availability in the cloud.

PR #430, opened 2022-10-14
How DNS request works after you type www.google.com

Kenneth1234:


Nonsi7:
The name of the website (URL) and the specific IP address it connects to are both kept in a database called DNS (Domain Name System). Because of this, DNS links the URL https://www.google.com to its specific IP address when we type it into our browser. DNS is a list of URLs and their corresponding IP addresses.

PR #439, opened 2022-11-04
Explained more on SPDY, to  aid understand more on it's functions

NwabuezeFranklin:


1f604:
SPDY is deprecated.

Sanious22:
alx

Gifftybabe:
SPDY (pronounced "speedy") is a networking protocol that was developed by Google in order to improve the performance of the HTTP protocol, which is the backbone of the World Wide Web. The basic idea behind SPDY is to reduce the amount of time it takes for a client to request and receive content from a server.

SPDY achieves this goal by introducing several key features, including:

- Multiplexing: SPDY allows multiple requests and responses to be sent and received simultaneously over a single connection, - - thereby reducing the number of connections needed to load a web page.

- Stream Prioritization: SPDY allows clients to specify which requests are more important than others, so that the server can prioritize its responses accordingly.

- Server Push: SPDY allows servers to push content to the client before it is explicitly requested, so that the client can start rendering the page faster.

- Header Compression: SPDY compresses HTTP headers before they are sent over the wire, thereby reducing the amount of data that needs to be transmitted.

- SSL/TLS encryption: SPDY requires the use of SSL/TLS encryption, which provides an additional layer of security for web traffic.

Overall, SPDY provides significant performance benefits for web applications, particularly for those that involve large numbers of small requests, such as AJAX-heavy websites. In 2015, the HTTP/2 standard was released, which incorporated many of the features of SPDY and has since replaced it as the preferred protocol for modern web applications.

Aminj66:
https://github.com/alex/what-happens-when/pull

PR #444, opened 2022-11-05
DNS Lookup

Tesfay-Teshome:
What Does DNS Lookup Mean?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,6 +39,11 @@ popular searches from the internet as a whole. As you are typing
 with each keypress. It may even suggest "google.com" before you finish typing
 it.
 
+DNS Lookup
+----------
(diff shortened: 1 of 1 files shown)


alextakele:
General Software Engineering interview Questions and answer

Majey:
The DNS server is like the contact app on our phones whose purpose is to match contact names with numbers. The DNS matches domain names to IP addresses. DNS lookup refers to the process of mapping these domain names to their respective web server IP addresses.

alextakele:
> The DNS server is like the contact app on our phones whose purpose is to match contact names with numbers. The DNS matches domain names to IP addresses. DNS lookup refers to the process of mapping these domain names to their respective web server IP addresses.

Good explanation

PR #445, opened 2022-11-05
Update README: update what happens when the g key is pressed

n1klaus:
Added more information on the processes that take a place in the CPU and the OS and some of the most popular algorithms used.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -25,17 +25,26 @@ Table of Contents
    :backlinks: none
    :local:
 
+The following sections explain the physical keyboard actions and the OS interrupts.
+
 The "g" key is pressed
 ----------------------
-The following sections explain the physical keyboard actions
-and the OS interrupts. When you press the key "g" the browser receives the
-event and the auto-complete functions kick in.
-Depending on your browser's algorithm and if you are in
-private/incognito mode or not various suggestions will be presented
-to you in the dropdown below the URL bar. Most of these algorithms sort
+When you press the key "g" in the browser, a switch closes and current 
+flows into a small chip in the keyboard. A scan code associated to the key 
+is sent to the CPU as a binary number. The OS then converts the binary number 
+into ASCII or Unicode character depending on the system's keyboard configuration. 
+The operating system then creates an event depending on the resulting character
+and sends the event to the application that was active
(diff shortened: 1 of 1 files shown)


wizleo:
The following sections explains the physical keyboard actions and the OS interrupts. When you press the key "g" the browser receives the event and the auto-complete functions kick in. Depending on your browser's algorithm and if you are in private/incognito mode or not various suggestions will be presented to you in the dropdown below the URL bar. Most of these algorithms sort and prioritize results based on search history, bookmarks, cookies, and popular searches from the internet as a whole. As you are typing "google.com" many blocks of code run and the suggestions will be refined with each keypress. It may even suggest "google.com" before you finish typing it.

i-be-k:
When the key 'g' is pressed, the first thing that happens is that your browser looks up in its memory to see if that website was visited before and the IP address is known. If it can't find the IP address for the URL requested then it asks your operating system to locate the web site. The first place your operating system is going to check for the address of the URL you specified is in the hosts file (/etc/hosts in Linux and Mac, c:\windows\system32\drivers\etc\hosts in Windows). If the URL is not found inside this file, then the OS will make a DNS request to find the IP Address of the web page. The first step is to ask the Resolver (or Internet Service Provider) server to look up in its cache to see if it knows the IP Address, if the Resolver does not know then it asks the root server to ask the .COM TLD (Top Level Domain) server - if your URL ends in .net then the TLD server would be .NET and so on - the TLD server will again check in its cache to see if the requested IP Address is there. If not, then it will have at least one of the authoritative name servers associated with that URL, and after going to the Name Server, it will return the IP Address associated with your URL. All this will be done in a matter of milliseconds.

SamanSurafel17:
The following sections explain the physical keyboard actions and the OS interrupts. When you press the key "g" the browser receives the event and the auto-complete functions kick in. Depending on your browser's algorithm and if you are in private/incognito mode or not various suggestions will be presented to you in the drop-down below the URL bar. Most of these algorithms sort and prioritize results based on search history, bookmarks, cookies, and popular searches from the internet as a whole. As you are typing "google.com" many blocks of code run and the suggestions will be refined with each key press. It may even suggest "google.com" before you finish typing it.

PR #515, opened 2023-01-08
minor readme fix

ofemjohn:
added a little text on the subject matter


BirukL137:
Some texts has been added

PR #525, opened 2023-01-09
added  meaning of DNS AND ARP TO README

favour-22:


favour-22:
meaning of dns and arp added

PR #533, opened 2023-01-11
Added DNS Explanation

daveeazi:
I added the DNS Information

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -209,6 +209,8 @@ Check HSTS list
 DNS lookup
 ----------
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -209,6 +209,8 @@ Check HSTS list
 DNS lookup
 ----------
 
+Domain Name System (DNS) is, in simple words, the technology that translates human-adapted, text-based domain names to machine-adapted, numerical-based IP. When you wish to visit google.com, you ought to type in the address as http://209.85.227.104 before you can have access to google, but because of this amazing thing called DNS, you don’t have to memorize dozens of IP addresses you visit frequently. You can compare DNS and caching (cache-ing) to the contacts app on your device. If you need to call your friends, you don’t have to know all of their phone numbers by heart because the app takes care of that; you simply have to know the name of the person you wish to call, as it is saved in the contacts app
```
k2chogori, line 212: * `in simple words` - no additional value
* `machine-adapted` - what does it mean? Do you have me precise explanations?
* `you ought to type in the address as http://209.85.227.104 before you can have access to google` - there is not explanations why?
* `but because of this amazing thing called DNS` - this is technical document, no value adding with this
* `and caching (cache-ing)`- are you explaining "cache" or DNS?
* `If you need to call your friends, you don’t have to know all of their phone numbers by heart because the app takes care of that; you simply have to know the name of the person you wish to call, as it is saved in the contacts app` - can you please make it more technical?

kennedypascal: used your information to explain what happens when you type google.com and press enter

kennedypascal (commented):
well understood

PR #559, opened 2023-01-28
Update README.rst

ternguy:
it has to do with the DNS in conjuction will the IP addresses pointing to the domain server is important


ternguy:
it has to do with the DNS in conjuction will the IP addresses pointing to the domain server.

ternguy:
this explain what happens when you type www.google.com in the browser url and press enter.

PR #569, opened 2023-01-29
About DNS

MuneneCalvin:


MuneneCalvin:
DNS request: The first step is to translate a domain name into an IP address. Your browser sends a request to the DNS server to resolve the domain name to an IP address.

PR #601, opened 2023-01-30
load balancing

calvean:
added how google manage traffic


KoxyG:
It is important to note that the Load balancer uses a  variety of algorithms to make certain decisions, such as round-robin, least connections, or IP hash.

PR #639, opened 2023-04-13
Explain the HTTP verbs listed

allenkays:
Added explanation for the HTTP verbs listed


allenkays:
Hoping the flow is okay

nellencr:
HTTP verbs are used in HTTP requests to communicate with web servers and perform various operations on resources. They form the backbone of the HTTP protocol and are essential for building modern web applications.

PR #643, opened 2023-04-13
load balancer

Pokkode:
added load balancers for high traffic websites & servers


Gr8Samuel:
HAproxy is a very famous load-balancer, and example of algorithms that we can use are the round-robin, which distributes the requests alternating between all the servers evenly and consequentially, or the least-connection, which distributes requests depending on the current server loads.

PR #658, opened 2023-04-14
Update on Is it a URL or a search term?

Anto236:
Update  for ease of understanding


greenfield01:
read through carefully

PR #665, opened 2023-04-14
what-happens-when

BusolaAM:
When a TCP sender detects a congestion event (i.e. a packet loss or delay),
it reduces the transmission rate to prevent further congestion.
The sender then gradually increases the rate again, monitoring the network
for congestion events and adjusting the rate accordingly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -405,6 +405,10 @@ will be dropped before they get to their final destination. The sender then has
 to decide how to react. The algorithm for this is called `TCP congestion
 control`_. This varies depending on the sender; the most common algorithms are
 `cubic`_ on newer operating systems and `New Reno`_ on almost all others.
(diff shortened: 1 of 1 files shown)


1923k:
Client-Server Communication

AniKanayo:
The TCP congestion-avoidance algorithm is the foundational block for congestion control in the internet.

Dagwal:
When a packet is lost, the congestion window size is set to 1. Then when we retransmit the packet, if we receive a cumulative ACK for a lot of subsequent packets, we can assume that the packet loss was not due to congestion, but because of some random factors.

Dagwal (commented):
When a packet is lost, the congestion window size is set to 1. Then when we retransmit the packet, if we receive a cumulative ACK for a lot of subsequent packets, we can assume that the packet loss was not due to congestion, but because of some random factors.

PR #684, opened 2023-04-15
Adding some file to new branch

ZahraJkhan:


Victor-Amonde:
What happens when you type a URL into your browser?
The process involves the browser, your computer’s operating system, your internet service provider, the server where you host the site, and services running on that server.

When you type and enter a URL like https://www.google.com/ in your browser, your browser has to figure out which server on the Internet is hosting the site. It does this by looking up the domain, google.com, to find the IP address. Domain names are much easier to remember then IP address.

The domain name system, or DNS helps our browser (and us) find servers on the Internet. We can do a DNS lookup to find the IP address of the server based on the domain name, google.com.

Below is the explanation of steps involved

Let’s break down the parts of this URL you typed in
https://www.google.com/

https://is the scheme. HTTPS stands for Hypertext Transfer Protocol Secure. This scheme tells the browser to make a connection to the server using Transport Layer Security, or TLS. TLS is an encryption protocol to secure communications over the Internet. With HTTPS, the data exchanged between your browser and the server is encrypted.
[google.com](https://www.google.com/) is the domain name of the site. It is the memorable address and points to a specific server’s IP address.

https://www.google.com/the-path-to/the-resource[,](https://www.google.com/the-path-to/the-resource,the-path-to) the-path-to is the path on the server to the requested resource, the-resource. It’s a way to organize your resources, whether they are static HTML, CSS, Javascript, or image files, or dynamically generated content. Common examples of paths areblog,posts,tags, orimages, each grouping different resources.

When you typed this URL into your browser,the-resource is the name of the resource on the website you want to view. Sometimes you’ll see this with a file extension like.html which indicates this is a static file on the server with HTML content. Without an extension, like this URL, it usually indicates the server generated this content.

Browser looks up IP address for the domain
After you’ve typed the URL into your browser and pressed enter, the browser needs to figure out which server on the Internet to connect to. To do that, it needs to look up the IP address of the server hosting the website using the domain you typed in. It does this using a DNS lookup.

Because DNS is complex and has to be fast, DNS data is cached at different layers between your browser and at various places across the Internet. Your browser checks its own cache, the operating system cache, a local network cache at your router, and a DNS server cache on your corporate network or at your internet service provider (ISP). If the browser cannot find the IP address at any of those cache layers, the DNS server on your corporate network or at your ISP does a recursive DNS lookup. A recursive DNS lookup asks multiple DNS servers around the Internet, which in turn ask more DNS servers for the DNS record until it is found. Once the browser gets the DNS record with the IP address, it’s time for it to find the server on the Internet and establish a connection.

Browser initiates TCP connection with the server
Using the public Internet routing infrastructure, packets from a client browser request get routed through the router, the ISP, through an internet exchange to switch ISPs or networks, all using transmission control protocol, more commonly known as TCP, to find the server with the IP address to connect to. But this is a very roundabout way to get there and it’s not efficient.

Instead, many sites use a content delivery network, or CDN, to cache static and dynamic content closer to the browser. A CDN is a globally distributed network of caching servers that improve the performance of your site or app (the origin) by bringing the content closer to your users. Once the browser finds the server on the Internet, it establishes a TCP connection with the server and if HTTPS is being used, a TLS handshake takes place to secure the communication. Once the browser has established a connection with the server, the next step is to send the HTTP request to get the resource, or the page.

Browser sends the HTTP request to the server
Now that the browser has a connection to the server, it follows the rules of communication for the HTTP(s) protocol. It starts with the browser sending an HTTP request to the server to request the contents of the page. The HTTP request contains a request line, headers (or metadata about the request), and a body. The request line contains information that the server can use to determine what the client (in this case, your browser) wants to do.

The request line for the URL request looks like this:

GET /the-resource HTTP/1.1

The request line tells the server that you want to GET resource at /the-resource and to communicate with HTTP/1.1.

Remember that HTTP verbs express the semantic intent of your request. You could also use the POST, PUT, or PATCH methods to send data to the server for storage, either to create new data or update existing data at the request path. The DELETE method would delete the resource at the given path. However, it’s important to know that servers don’t have to support all verbs. A server could respond with a200 OK status to a DELETE method and not do anything with the resource.
The next part of the request is the request headers. Headers pass extra information along from the client that help route the request, indicate what type of client is making the request (the user agent), and can be used for supporting A/B testing, blue/green deployments, and canary deployments.

The last part of the request is the body. The body is (usually) empty for a GET request like ours. For a request that manipulates resources, like POST, PUT, or PATCH, the body will contain the data from the client to create or update.
The request body can be in different formats and the server understands the format based on a request header,Content-Type. Once the server has received the request from the client, the server processes it and sends back a response.

Server processes request and sends back a response
The server takes the request and based on the info in the request line, headers, and body, decides how to process the request. For the request, GET /the-resource/ HTTP/1.1, the server gets the content at this path, constructs the response and sends it back to the client. The status line contains both the HTTP version and a numeric and text representation of the status of the request

The browser considers a status code in the 200s to be successful. The response was 200, so the browser knows to render the response.
Resources can be static files, either text (i.eindex.html) or non-text data (i.e.logo.img). Browsers aren’t always requesting static files, though. Often, these are dynamic resources generated at the time of the request and there is no file associated with the resource. In fact, in this request, that’s exactly what is happening. There is no file the-resource, but the server still knows how to process the request. The server generates a dynamic resource, building the HTML from fragments or templates and combining it with dynamic data to send back in the response, as text, so the browser can render the page.
Next how the browser handles the response.

Browser renders the content
Once the browser has received the response from the server, it inspects the response headers for information on how to render the resource. TheContent-Type header above tells the browser it received an HTML resource in the response body. Lucky for us, the browser knows what to do with HTML!
The first GET request returns HTML, the structure of the page. But if you inspect the HTML of the page (or any web page) with your browser’s dev tools, you’ll see it references other Javascript, CSS, image resources and requests additional data in order to render the web page as designed.
As the browser is parsing and rendering the HTML, it is making additional requests to get Javascript, CSS, images, and data.

PR #687, opened 2023-04-15
Wrote a blog post about this

davidudo:
I wrote a blog about this topic and placed a link in a file named 0-blog_post_link. You should check it out.

### DIFF
--- a/0-blog_post_link
+++ b/0-blog_post_link
@@ -1,0 +1,1 @@
+https://medium.com/@udodavid46.ud/what-happens-when-you-type-google-com-in-your-browser-and-press-enter-5319f02368a9


Favour919:
### What happens when you type google.com in your browser and press Enter

Please follow the below steps for the entire process

No.1
When typing https://www.google.com/ on the web browser, the browser performs a Domain Name System(DNS) lookup to find the IP address associated with “google.com”.

What is a DNS?

In accordance to AWS, DNS translates human readable domain names(for example, [www.amazon.com](http://www.amazon.com/) ) to machine readable address(for example, 192.0.2.44).
Every device in the internet like smartphone, laptop, desktop has its own unique number that is used to identify it and this number is called IP address. So the communication of all the device is facilitated via the IP addresses. Since, the internet has a very huge of devices connected, it is not easy for humans to remember or store all this IP addresses in their memory for every communication that can be made.
The solution to this matter is the DNS where by a human can type the domain name of the server (google.com) which still be correct as it is mapped to its IP address. So this involves sending request to DNS server for translating the domain name into an IP address.

No.2
Once the client browser has obtained the IP address of the server hosting “google.com”, it will then initiate a Transmission Control Protocol(TCP) connection to that server.

No.3
At this stage client browser will now sent DNS request to the web server for requesting the home page of the domain name (google.com).

No.4
Upon receiving of the DNS request by a web server, it will then return back a HTTPS response containing the homepage of the domain (google.com) and all with the associated information required. Please note that for the security of the connection between the client and server browser, the requests and response from both would be either secured or insecure. Secure connection would have (HTTPS) to show the presence of the SSL certificate while the insecure would have (HTTP) to show the absence of SSL certificate and the connection is not trust.

No.5
Now the client browser will show homepage of the domain (google.com).

No.6
Lastly, now since a client browser has displayed the homepage, depending on what activity done by a user, the web server would the authenticate the user to the database to verify so that the required action can be done.

In all this scenario, in a case when there is too much traffic populated the web server, the would be a standby server which would automatically share the traffic with exiting server avoid overload of existing server as to improve the efficiency. This is called Load-balancing.

PR #690, opened 2023-04-15
Added an explanation on Load balancer

callmhejerry:


Olatundeawo:
A load balancer can be used for efficiently distribution of network from the client to a group of backend server, the load balancer also ensure high availability and reliability by sending requests only to servers that are online

PR #716, opened 2023-04-16
What happens when...

Ashiegbu:
What happens when you type https://www.google.com and press the 'Enter' key.


sharon5555:
When you type "www.google.com" in your browser and press enter, your computer sends a request to a server at Google's IP address. The server responds by sending back the HTML code for the Google search page, which your browser renders and displays on your screen.

PR #732, opened 2023-04-16
Added to the Dns Look up

Jorshstar:


cedarworth:
A DNS lookup, in a general sense, is the process by which a DNS record is returned from a DNS server. This is like looking up a phone number in a phone book — — that is why it is referred to as a “lookup”. Interconnected computers, servers and smart phones need to know how to translate the email addresses and domain names people use into meaningful numerical addresses. A DNS lookup performs this function.

PR #735, opened 2023-04-16
Added load balancer

Uthmanduro:


siluye:
HTTP
A Hyper Text Transfer Protocol is a protocol for fetching resources such as HTML documents. It is the foundation of any data exchange on the Web.

PR #751, opened 2023-04-16
Create FrancisOkolo

FrancisOkolo:
googlecom_diagram


FrancisOkolo:
WHAT HAPPENS WHEN YOU TYPE google.com IN YOUR BROWSER AND PRESS ENTER

Browsers are special applications built to accomplish complex web processes behind the hood and present neat results which end users can appreciate.

Domain names are words which represent IP addresses of remote servers. These remote servers contain web pages, files and applications which the user requests for. The browser works with a Domain name server (DNS). the DNS maps the domain name to the IP address of the server hosting the file or application needed by the user. Once the DNS gets the IP address, it sends this address to the browser.

The browser begins to prepare an IP/TCP to connect with the remote server. Depending on the type of resource requested by the user, the client uses the appropriate protocol to communicate with the remote server. In the case of google.com, the hypertext transfer protocol secure (HTTPS). The HTTPS works in the application layer of the TCP/IP model and effectively secures the data packets sent through the network to prevent leakage or deliberate unauthorized  attempts to access information contained in the data packets.
HTTPS works using a secure socket layer (SSL). SSL simply encrypts the data package so that unauthorized users cannot understand the information contained in the packet. Only the intended client can decode the data and it does so by means of a certificate generated by the HTTPS installed in the remote server

When connection between the client and server is established, the transport layer begins to break the requests into units known as packets. These packets can be sent separately through different network routes to ease traffic. This traffic is filtered by a network security device called a Firewall. A firewall has the responsibility of checking the packets and ensuring that only appropriate and certified packets are passed through the network. A firewall’s main purpose is to allow non-threatening traffic in and to keep dangerous traffic out.

Large scale search engines like google.com use several servers for efficient management of traffic. Since there can be millions of clients accessing the website at a time. Traffic management is done by a special server called Load Balancer. The load balancer works by distributing requests to the different servers using a reliable technique that ensures smooth traffic.

The web server is a system software which hosts webpages, files such as audio, video, images or even applications needed by clients. These web servers contain the page requested by the client so when the load balancer sends these request packets to the server, it delivers it to the load balancer and the loader sends the response to the client.

In the case where the client requests for a resource with functionality, e.g  google docs or google meet. This request is handled by a special server known as Application server.

The web server alone cannot handle the storage required to store all the resources in google’s archive. So a special system is deployed to handle these mammoth shelves of documents, images, videos and audio. This special system is called a Database.

PR #752, opened 2023-04-16
What happen when you type Google.com

Masterprecie:


elahbazzy:
when you type for example https://google.com/ into your browser and press Enter, your browser sends a DNS request to translate the domain name [www.google.com](http://www.google.com/) into an IP address. It then establishes a connection with the server using TCP/IP which passes through a firewall, and initiates a secure connection using HTTPS/SSL. The Load balancer selects a server which sends the requested resources to your browser. In case of further processing, the web server passes the request to an Application server which will access the database to retrieve data. Finally, the Application Server generates a response, which is sent to the Web Server and then to your browser. All this happen within few seconds.

ToluDaniels:
when you type for example https://google.com/ into your browser and press Enter, your browser sends a DNS request to translate the domain name [www.google.com](http://www.google.com/) into an IP address. It then establishes a connection with the server using TCP/IP which passes through a firewall, and initiates a secure connection using HTTPS/SSL. The Load balancer selects a server which sends the requested resources to your browser. In case of further processing, the web server passes the request to an Application server which will access the database to retrieve data.

Thermiprettie:
When you type https://www.google.com in your browser, your browser sends this request to the DNS server which will return the IP address of the web server that hosts the Google website. Once the IP address is known, the browser establishes a Transmission Control Protocol (TCP) connection with the server. This request will [pass through a firewall before it gets to the web server to filter out any malicious traffic. this then initiates a secure connection through the HTTPS

PR #763, opened 2023-04-16
Update README.rst

serena0012:


serena0012:
I made changes to the topic "What happens when you type the URL, google.com to your browser and hit enter"

serena0012:
Many of us are familiar with using the search engine, Google, to find information on the Internet.  Have you ever wondered what happens in the background when you type the URL(Uniform Resource Locator) [www.google.com](http://www.google.com/) on your browser and hit enter?  A whole lot of complex processes take place behind the scene before the requested web page is displayed on your screen. These processes take place within seconds you can hardly take notice of them.
I am going to take you through the journey of simplifying every process. So sit back, relax, and enjoy!
I will begin by defining some terminologies to help you get a good grasp of the whole thing.
Web browser: This is an application that you use to gain access to a website (e.g Google Chrome, Mozilla, etc)
Webpage: A webpage is a document that is shown on your computer screen
Website: A collection of webpages that are grouped in different fashions.
Web Server: A computer that hosts a website on the internet
Having done that, now let's dive deep into what happens when you type the URL, [www.google.com](http://www.google.com/) in your browser and press enter.
 Your browser first searches the DNS (Domain Name System), which is like an address book for websites, for the website's IP address before it can retrieve the website. The DNS, as I  earlier said, works like an address book that matches websites also called domain names to their corresponding unique IP addresses. For instance, [www.google.com](http://www.google.com/) will be matched to the IP address, 142.251.32.46. Similarly, [www.jw.org](http://www.jw.org/) will be matched to the IP address, 23.4.176.34, and so on.
So what happens next? Your browser establishes a connection with the website’s server through a communication protocol called TCP(Transmission Control Protocol).
TCP protocol ensures reliable and ordered data delivery between two devices over the internet. It works together with the Internet Protocol (IP), which is responsible for routing data between devices.
But hey! We are not done yet, there are more processes involved in this. Before the connection is established there is one more thing. Your browser has to pass through something called Firewall. Firewall serves as a security mechanism that checks or filters network traffic. It can either let it pass or block it, according to some specified rules. Think of it as a bouncer in a nightclub, he checks or searches you to see if you satisfy the criteria to gain access to the club according to the rule and regulations guiding the club. If you pass the security checks you are allowed in, otherwise, you are bounced back!
For a website using HTTPS(HyperText Transfer Protocol Secure)/SSL(Secure Socket Layer) for security purposes, your browser will establish a secure connection with the server to ensure that your data is encrypted(coded language) and cannot be accessed by unauthorized persons(scammers or hackers). This is especially important for protecting sensitive information such as passwords, credit card numbers, financial transactions, and other personal data transmitted over the Internet. So once a secure HTTPS connection has been established data will begin to be exchanged between the browser and the server. This is why you should look out for a padlock symbol on the URL address bar of your browser. 
The padlock symbol indicates that the website uses a secure connection, typically through HTTPS, a protocol for securely transmitting data over the internet. When you see a padlock symbol in your browser's address bar, it means that any data you exchange with the website is encrypted and cannot be intercepted or read by anyone else. Of course, it’s for the website’s security. However, it's worth noting that not all websites that use HTTPS will display the padlock symbol, and vice versa. Some browsers also display additional information when you click on the padlock symbol, such as the website's security certificate and the level of encryption being used.
Wait a minute, we’re almost done! 
Sometimes many websites you try to access are hosted on many servers. Why? It’s to handle request overloads. Imagine millions of users trying to access a website hosted on a single server. This will crash the server. This is the reason why there is what we call a load balancer server. The load balancer acts as a traffic cop, routing incoming requests to the most appropriate server based on factors such as server load and geographic location.
Finally, once your browser has made it past those steps, the web server can now receive the request and retrieve the requested web page from the application server, which in turn retrieves information from the database to generate dynamic content that is displayed on the web page for you to see.
In summary, this is what happens when you type the URL, [www.google.com](http://www.google.com/) in your browser and hit enter! As we have seen the process is complex and involves multiple steps. Each step plays a vital role in ensuring that the website is displayed quickly and securely in your browser.

PR #767, opened 2023-04-16
Create Load-balancer

itzcephas:


DELMUS1M:
In this post we will be learning what happens behind the scenes when we type a url ([www.google.com](http://www.google.com/)) in our web browser and how it reaches the website you want within a matter of few seconds…
When we type https://www.google.com/ into a browser window and hit enter, it magically goes to the site right? Not really, it’s not that simple. There are quite a few complicated steps involved that lead us to the destination we typed.
Way before even typing the URL, there are few things we need to access a browser (FireFox, Internet Browser, Google Chrome etc.). We need a functioning PC or Desktop with a Keyboard, internet (WIFI or mobile data).
If we have all above mentioned essentials, we start by;
i) Switching on the PC or Desktop.
ii) Type in your password if you activated one.
iii) Open any browser of your choice then type in the address you want, Here we would be focusing on https://www.google.com/

After typing the address, hit enter. Once you hit enter, a series of things happen within those few seconds that the website takes to load. Now let's comb through what happens behind closed doors and take a deeper look at how the website loading process works. First of, let's start with;

DNS Request

The first step is a DNS (Domain Name System) request. Your browser sends a request to your IP's (Internet Provider) DNS servers to resolve the domain name "[www.google.com](http://www.google.com/)" to an IP address(eg, 192.0.2.44). This IP address will then be used to communicate with the web server(software that delivers web pages).

TCP/IP

Once your browser knows the IP address of the web server, it establishes a TCP/IP (Transmission Control Protocol/Internet Protocol) connection to it. TCP/IP is a suite of communication protocols used to establish network connections and transfer data between devices.
TCP defines how applications create communications channels, it also manages how a message is broken down into packets before being transmitted over the internet and reassembled at the destination.
IP defines how to address and route each packets to ensure it reaches the right destination.

Firewall

Before establishing the connection, the Firewall comes into play. A firewall is a network security system that monitors and controls incoming and outgoing network traffic. The firewall checks the connection request and blocks it if it is suspicious or allowing data packets.

HTTPS/SSL

If the Firewall allows the connection to go through, the browser and the web server proceed with a secured process that uses the HTTPS (HyperText Transfer Protocol Secure) protocol. This process involves the exchange of SSL (Secure Socket Layer) certificates that help to encrypt the information being exchanged. HTTPS appears in the URL when a website is secured by an SSL certificate. The certificate includes the issuing of authority and the corporate name of the website owner.

Load-Balancer

Once the connection is established, the request is passed to a load balancer. A Load-balancer is a device that distributes network traffic to multiple web servers to improve performance and prevent any single server from being overloaded. Load balancers help to distribute requests optimally for faster response time. They are used to increase capacity and reliability of applications.

Web Server

After leaving the Load-balancer, the request reaches a web server. A web server is a computer program that responds to HTTP requests from clients such as web browsers. The web server processes the request and sends back a response. This response may include HTML (Hypertext Markup Language) code, CSS (Cascading Style Sheets), JavaScript, or any other resources needed to display the webpage correctly.

Application Server

If the response includes any dynamic content, such as information that needs to be retrieved from a database, it will be passed to an Application server. An Application server is a program that provides an environment where applications can run, and it processes requests from web servers to retrieve any necessary data.

Database

If the Application server requires access to a database, it will communicate with the database server to retrieve the necessary data. The database server is a computer program that stores and retrieves data when requested by an application program.

There you have it! These are the typical steps that occur when you type https://www.google.com/ in your browser and press Enter. Even though they happen in the background, these steps are essential to ensure a seamless browsing experience.

PR #794, opened 2023-04-17
contributing to the DNS lookup content

ayindeHassan:
made two additional point in the DNS lookup content.


Jakmike:
ABOUT DNS lookup

This test will list DNS records for a domain in priority order. The DNS lookup is done directly against the domain's authoritative name server, so changes to DNS Records should show up instantly. By default, the DNS lookup tool will return an IP address if you give it a name (e.g. www.example.com)

PR #811, opened 2023-04-17
Update README.rst

Emma0147:
i created an article on what happens when you type https://www.google.com/ in your browser and press Enter


Emma0147:
i wrote an artice on what happens when you type https://www.google.com/ in your browser and press Enter

PR #819, opened 2023-04-18
Create what happens

Gr8Samuel:


miiieagle:
Step1: DNS Request

The first thing that happens is that your computer sends a request to a Domain Name System (DNS) server to resolve the domain name "www.google.com" into an IP address.


Step2: TCP/IP

Once the DNS server responds with the IP address associated with the domain name, your computer uses the TCP/IP protocol to establish a connection with the web server.


Step3: Firewall

Before the data can reach the web server, it must pass through a firewall. The firewall checks the data packets to make sure they meet the necessary security requirements.


Step4: HTTPS/SSL

Once a connection is established between your computer and the web server, the next step is to establish a secure connection using HTTPS/SSL. This encryption ensures that sensitive data, such as login credentials or credit card information, cannot be intercepted by third parties.


Step5: Load-balancer

Many popular websites, including Google, use load-balancers to distribute incoming traffic across multiple servers. Load-balancers help to ensure that no single server becomes overloaded with traffic, which can cause the website to slow down or even crash.


Step6:Web server

The load-balancer then forwards the encrypted data packets to a web server. The web server is responsible for handling HTTP requests from clients and sending HTTP responses back to them. When the web server receives the encrypted data packets, it decrypts them using the SSL/TLS encryption and forwards the HTTP request to the appropriate application server.


Step7:Application server

In addition to serving static files like HTML and CSS, many websites rely on application servers to generate dynamic content. Application servers are responsible for running the code that generates dynamic content, such as search results or personalized recommendations. These servers may also interact with databases to retrieve and store information.


Step8:Database

Finally, many websites rely on databases to store and retrieve information. Databases are used to store user information, product catalogs, and other types of data that need to be accessed quickly and efficiently. Application servers interact with databases to retrieve and store information as needed.

PR #826, opened 2023-04-19
Add conclusion

Seamas007:
Please Add conclusion to README.rst


amirgambo:
published an article explaining how the client request is sent to the local DNS, to the DNS resolver, and the rest of the steps necessary to get the server IP address thereafter sending the server IP to the client.
In consideration a situation where a firewall and a load balancer on the server.

PR #832, opened 2023-04-21
https request

lucydevonne:


Monehinfeyi:
**## What happens when you type ''www.google.com'' in the address bar and press enter?**
It seems magic when an internet user types "https://www.google.com" in a browser and presses Enter, in a twinkle of an eye the needed information appears. It may seem like a simple action, but there are actually several complex processes that occur behind the scenes to make it happen. In this timely post, I will walk you through the journey of a web request, step by step.

  ###   DNS Request

The first thing that happens when an internet user types a  Universal Resources Locator(URL) into a browser is a **Domain Name System** (DNS translates human-readable domain names [for example, www.facebook.com] to machine-readable IP addresses [for example, 192.0.2.44]) request. The DNS is like the phonebook of the internet, matching domain names to IP addresses. When the user type "https://www.google.com" into a browser, the client (computer) sends a request to a DNS server to look up the IP address associated with that domain name. If the DNS server has the IP address in its cache, it returns it immediately. Otherwise, it queries other DNS servers until it finds the correct IP address. Once the IP address is found, it is returned to the browser.  
Read more about DNS https://en.wikipedia.org/wiki/Domain_Name_System

   ###  TCP/IP

Once the browser has the IP address, it needs to establish a connection with the web server that hosts the website. This is where the **Transmission Control Protocol (TCP) and Internet Protocol (IP)** come in. TCP is responsible for breaking up data into packets and ensuring that they are received in the correct order, while IP is **responsible** for routing the packets to their destination. The browser uses TCP/IP to establish a connection with the web server using the IP address obtained from the DNS server.
More about TCP/IP https://www.techtarget.com/searchnetworking/definition/TCP-IP
  
###  Firewall
Before the request reaches the web server, it may have to pass through a firewall. A firewall is a network security system that monitors and controls the incoming and outgoing network traffic based on predetermined security rules. It helps to prevent unauthorized access to a network and protect against attacks. If the request is deemed safe, it is allowed through the firewall and sent to the web server.
More about firewalls https://www.simplilearn.com/tutorials/cyber-security-tutorial/what-is-firewall

###     HTTPS/SSL

Once the request reaches the web server, it needs to be encrypted using **HTTPS (Hypertext Transfer Protocol Secure) with SSL (Secure Sockets Layer) or TLS (Transport Layer Security)** protocols. This means that any data transmitted between the browser and the web server is encrypted and secure. An internet user can tell if a website is using HTTPS by examining the padlock icon in the address bar of the browser. https://en.wikipedia.org/wiki/HTTPS

   ###  Load-balancer

Many high-traffic websites like Wikipedia, Craigslist, Fandom, IMDb, Google, and others use load balancers to distribute incoming requests across multiple web servers. **A load-balancer** can help to ensure that no single web server becomes overloaded with too many requests, helps to improve the overall performance and reliability of the website, and aids quick response during requests.  https://www.nginx.com/resources/glossary/load-balancing/

    ### Web server

The web server is primarily used to process and manage Hypertext Transfer Protocol **(HTTP) / Hypertext Transfer Protocol Secure (HTTPS)** requests and responses from the client system. Once the request reaches the web server, it is processed by the web server software. The web server retrieves the requested web page and sends it back to the browser as a response. The web server will execute any script or program required to generate the web page content dynamically. https://en.wikipedia.org/wiki/Web_server

    ### Application Server

In some cases, the web server may need to communicate with an application server to generate the web page content. An application server is a software framework that provides a platform for running web applications. It handles tasks such as authentication, session management, and database access. The web server communicates with the application server to generate the dynamic content of the web page. https://en.wikipedia.org/wiki/Application_server

  ###   Database

**Finally**, a database is an organized collection of structured information, or data, typically stored electronically in a computer system. The web server may need to retrieve data from a database to generate the web page content. The database stores the data required by the web application, such as sales transactions, customer data, financials, product information, blog posts,  etc. The web server communicates with the database to retrieve the required data and generate the web page content. https://en.wikipedia.org/wiki/Database

**In** conclusion, typing "https://www.google.com" into a browser and pressing Enter triggers a complex series of processes that involve DNS requests, TCP/IP connections, firewalls, HTTPS/SSL encryption, load-balancers, database, and others.

PR #845, opened 2023-04-23
Edited DNS lookup section

PeterManda19:


Monehinfeyi:
**### What happens when the "g" key is been typed in a web browser on a physical keyboard?**
When the "g" key is typed in a web browser on a physical keyboard, the key press generates a keyboard event, which is sent to the web browser. The web browser then processes the event and sends it to the appropriate part of the web page, such as a search box.

If the user is typing in a search box, for example, the "g" key press will trigger the input of the character "g" into the search box. The web browser will display the character on the screen and may also offer auto-complete suggestions based on the user's previous search history or commonly used search terms.

If the user has configured their web browser to use a search engine as the default search provider, the browser may also initiate a search request using the text entered into the search box, which would include the "g" character if it was typed. The search engine would then return results based on the query, which the browser would display to the user.

PR #855, opened 2023-05-01
i just work on window server

Yettyola:
This is the task on when i fixing the ALX code base for contribution 
window server

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -677,6 +677,12 @@ GPU Rendering
 
 Window Server
 -------------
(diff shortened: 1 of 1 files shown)


Ibiniyi:
**Different Between Server and Web Server**
A web server is a specific type of server that is designed to handle HTTP requests and responses. It is responsible for responding to requests from web browsers and serving web pages. On the other hand, a server is a more general term that can refer to any type of computer or device that provides services or resources to other devices or clients over a network. This can include file servers, print servers, database servers, email servers, and more

PR #868, opened 2023-05-12
Asakipaam's contribution

Asakippam-Simon:
Asakipaam's explanation of the process that occurs when a user searches for a URL, like https://www.google.com/, in a web browser such as Chrome.


Ibiniyi:
A web server is a specific type of server that is designed to handle HTTP requests and responses. It is responsible for responding to requests from web browsers and serving web pages. On the other hand, a server is a more general term that can refer to any type of computer or device that provides services or resources to other devices or clients over a network. This can include file servers, print servers, database servers, email servers, and more****

Ibiniyi:
Different Between Web server and Server 
A web server is a specific type of server that is designed to handle HTTP requests and responses. It is responsible for responding to requests from web browsers and serving web pages. On the other hand, a server is a more general term that can refer to any type of computer or device that provides services or resources to other devices or clients over a network. This can include file servers, print servers, database servers, email servers, and more

PR #879, opened 2023-05-13
dns

lizodus:


Attamakevin:
When you type a domain name eg. "google.com" into your browser and hit Enter, your browser first checks its own cache to see if it has already looked up the IP address for that domain name recently. If it has, it can skip the rest of the DNS resolution process and use the cached IP address to connect to the server.
If the IP address isn't in the cache, your browser sends a DNS query to your local DNS resolver, which is typically provided by your Internet Service Provider (ISP). The DNS resolver checks its own cache to see if it has the IP address for "google.com". If it does, it sends it back to your browser, which can then connect to the server using that IP address.
If the IP address isn't in the local resolver's cache, it sends a recursive query to the root DNS servers. These are the servers at the top of the DNS hierarchy, and they maintain information about the DNS servers responsible for each top-level domain (such as ".com", ".org", etc.).
The root DNS server responds to the query with a referral to the DNS server responsible for the ".com" top-level domain. Your local resolver then sends another query to that server, which responds with a referral to the DNS server responsible for "google.com".
Finally, your local resolver sends a query to the "google.com" DNS server, which responds with the IP address for the Google server that hosts the "google.com" website. This IP address is sent back to your browser, which can then connect to the server.

PR #892, opened 2023-05-14
What happens when g is press

Rufai5:
My first Branch resolving What happen when "g" is press on the browser.


pyshola:
The browser sort list of domain name using likely priority queue to get recent or highest visited webpage to complete the url you want to enter.

PR #900, opened 2023-05-14
add details about load balancer and firewall

Almiviolad:
I added a brief detail about the load balancer and firewall. My little contribution


solomongegziabher:
A load balancer is a device that distributes network traffic across multiple servers to ensure that no single server is overwhelmed with traffic . Load balancers are used to improve the performance and reliability of websites and applications by distributing traffic across multiple servers and A firewall is a network security system that monitors and controls incoming and outgoing network traffic based on predetermined security rules . Firewalls are used to establish a barrier between an internal network and the internet to prevent unauthorized access to the internal network .

PR #904, opened 2023-05-14
My contribution to what-happens-when_github_answer #1

Etomu:


Engmbakri:
explanation of what happens when you type "https://www.google.com/" in your browser and press Enter. The explanation covers each step of the process, including DNS resolution, establishment of a secure connection using HTTPS, traversal through a firewall, load balancing, request handling by the web server, dynamic content generation by the application server, and data retrieval from the database. Additionally, a schema illustrating the flow of the request has been included to visually represent the sequence of actions involved in serving the requested web page. This addition aims to provide readers with a thorough understanding of the intricate journey of a URL request, enhancing the repository's value as a resource for learning about internet infrastructure and request handling processes.

PR #923, opened 2023-05-14
What happens ..

Avwerosuo25:
What happens...

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -689,6 +689,8 @@ the Google homepage. Scripts can cause additional network requests to be
 performed, as well as modify the page or its layout, causing another round of
 page rendering and painting.
 
+Add Images
+For better illustrations and understanding
 .. _`Creative Commons Zero`: https://creativecommons.org/publicdomain/zero/1.0/
 .. _`"CSS lexical and syntax grammar"`: http://www.w3.org/TR/CSS2/grammar.html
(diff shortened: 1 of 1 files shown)


E-buez:
What happens when you type [google.com](http://google.com/) in your browser

Upon typing in "[google.com](http://google.com/)" and pressing Enter, the browser takes a number of steps which can be outlined as follows:
1. Resolve IP address of the URL via DNS.
2. Generate an HTTP request with headers (accept, user-agent, cookie, etc).
3. Open an HTTP connection to the resolved IP address.
4. Send the request to the server.
5. Receive the response from the server.
6. Parse response headers.
7. Depending on the response headers, perform additional operations.
8. Decompress the response body if it's compressed (e.g. gzipped).
9. Parse the HTML code inside the response body.
10. Resolve any additional resources (images, stylesheets, scripts, etc).
11. Start loading those resources via their URLs using the same steps.
12. Render the HTML as soon as required resources are loaded, continue loading remaining resources in background.
13. When all the resources are loaded, close the HTTP connection.
https://www.linkedin.com/posts/ehujuo-chibueze-1200251b0_what-activity-7063645965420830720-nwAk?utm_source=share&utm_medium=member_desktop
https://imgur.com/a/bA0kKTS

PR #950, opened 2023-05-20
Add contribution to what happens when you type google.com and press e…

itsmraga-hub:
- DNS request
- TCP/IP
- Firewall
- HTTPS/SSL
- Load-balancer
- Web server
- Application server
- Database

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -663,6 +663,28 @@ Page Rendering
   via Direct3D/OpenGL. The GPU command buffer(s) are flushed to the GPU for
   asynchronous rendering and the frame is sent to the window server.
 
+you type google.com in your browser and press Enter
+---------------------------------------------------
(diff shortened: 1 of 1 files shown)


Mondabutu:
When you have point your browser at a URL like https://www.google.com,  and  make a search, the browser look at the domain name google.com to find the address and figure out which google server is hosting the  searched file. The server hosting the file contains four numbered parts referred to as IP address.

G4GIFTEDHANDS:
The process starts with a DNS request, which translates the domain name "google.com" into an IP address. The request is sent to a DNS server, which returns the IP address of one or more servers that host the website. The browser then establishes a TCP/IP connection with the web server, which sends back an HTTP response containing the web page's HTML code.

PR #993, opened 2023-07-14
Update README.rst and added content on caching

Revelation-coder:
Added a paragraph about how browsers handle requests faster by caching


Revelation-coder:
caching

PR #999, opened 2023-07-14
TLS handshake

olecko:
This is further contribution to TLS Handshake


LevittC17:
More on TCP Handshake: To initiate a TCP connection, your browser sends a SYN (synchronize) packet to the IP address obtained from the DNS resolver. This packet contains various details, including source and destination ports and sequence numbers.

PR #1002, opened 2023-07-14
Create What happens when you type any url in your browser and press e…

OmobaVII:
…nter

### DIFF
--- a/what-happens_when_you_type_google.com_or_any_other_url_in_your_browser_and_press_enter
+++ b/what-happens_when_you_type_google.com_or_any_other_url_in_your_browser_and_press_enter
@@ -1,0 +1,38 @@
+
+Ever wondered about the intricate processes that occur behind the scenes when you type a website address into your browser and hit Enter? The journey of a web request involves a multitude of components working together seamlessly. In this blog post, we will dive into the details of what happens when you type "google.com" in your browser and press Enter, covering DNS request, TCP/IP, firewalls, HTTPS/SSL, load-balancers, web servers, application servers, and databases.
+
+1. DNS Request:
+
+The process commences with a Domain Name System (DNS) request. Your browser needs to translate the human-readable domain name "google.com" into an IP address. It consults DNS servers to obtain the corresponding IP address. If the IP address is not cached locally, the request is sent to DNS servers to perform the resolution.
+
+2. TCP/IP:
+
+With the IP address acquired, the browser establishes a connection using the Transmission Control Protocol (TCP) and the Internet Protocol (IP). TCP ensures reliable data transmission by breaking it into packets and reassembling them upon arrival at the destination.
+
+3. Firewalls:
+
+Firewalls play a crucial role in network security. Before accessing the web server, your request may encounter a firewall. Firewalls examine incoming and outgoing network traffic, implementing security policies to filter potentially malicious requests and protect against unauthorized access.
+
+4. HTTPS/SSL:
+
+To ensure secure communication, many websites employ Hypertext Transfer Protocol Secure (HTTPS) in conjunction with the Secure Sockets Layer (SSL) or Transport Layer Security (TLS) protocols. This encryption safeguards data transmission, protecting it from interception and manipulation. Your browser and the web server engage in an SSL/TLS handshake, establishing a secure connection.
+
+5. Load-Balancer:
+
+High-traffic websites often employ load-balancers to distribute incoming requests across multiple servers. Load-balancers enhance performance and ensure that no single server becomes overwhelmed. They intelligently distribute requests based on factors such as server availability, response times, and resource utilization.
+
+6. Web Server:
+
+The web server is responsible for handling HTTP requests. Upon receiving your request, the web server processes it, retrieves the requested resources, and generates an HTTP response. It may also execute server-side scripts, such as PHP or Python, to generate dynamic content.
+
+7. Application Server:
+
+In certain cases, the web server forwards requests requiring dynamic processing to an application server. Application servers execute server-side scripts and interact with databases and other resources to generate personalized responses. They enable the execution of complex business logic and facilitate scalability.
+
+8. Database:
+
+Many websites rely on databases to store and retrieve data. The application server communicates with the database to fetch relevant information needed to generate a response. Databases, such as MySQL or PostgreSQL, manage structured data efficiently and enable persistent storage of website content, user data, and more.
+
+Conclusion:
+
+The process of loading a webpage involves a harmonious collaboration among various components. From DNS requests translating domain names into IP addresses to establishing secure connections, passing through firewalls, distributing requests with load-balancers, processing on web and application servers, and interacting with databases, each step contributes to the successful delivery of a web page. Understanding this intricate process provides a glimpse into the impressive technology that powers our browsing experiences and enables the seamless retrieval of information from the internet.


adexoludare:
What Happens when you type [www.google.com](http://www.google.com/)
In these present world, almost everybody both young and old knows how to get information from the internet. Computing has made life easier as you can learn anything, connect with your families and friends on social media, watch news or movies online and so on with aid of internet. All you need to do is to:
Take up your device; mobile phone, tablet, your personal computer(PC). 
Open your favorite web browser.
In your address bar, type in the word/web addressee you want
hit enter key button on your keyboard.
Your browser display the result of your search back to you and will have the inflammation right before you. But the information not just come. What happen behind the scene ? 

When the words typed in is not a valid domain name, the browser proceeds to feed the text given in the address bar to the browser’s default web search engine. In most cases the URL has a special piece of text appended to it to tell the search engine that it came from a particular browser’s URL bar. But if valid domain name is typed in, the web browser makes the request to access the site and the request is sent to what’s known as a Domain Name System Resolver or just resolver. 

A Domain Name System can be liken to contact list on your mobile phone where you save the phone number of your friends when ever you want to call any of them, you just typed his/her name the contact display the phone number you make your call. in this illustration, contact name is synonymous to domain name while contact phone number to IP address. Also DNS can be thought of as an address book that catalogs a website’s name, or domain, with its respective IP address. 

This is useful because we don’t have to concern ourselves with typing in the exact IP address for the website address we would like to visit. We can simply type in the name of the website and the DNS resolver will attempt to locate the name of the site and the IP address that corresponds to it. These resolver that help make these translations for us are commonly found in large institutional networks such as your local internet service provider.

The next step is the actual transfer of information between your computer and the computer it’s looking to connect with. This is achieved with the aid of network protocol.
 
What is a network protocol?
A network protocol is a set of established rules that specify how to format, send and receive data so that computer network endpoints, including computers, servers, routers and virtual machines, can communicate despite differences in their underlying infrastructures, designs or standards.

To successfully send and receive information, devices on both sides of a communication exchange must accept and follow protocol conventions. In networking, support for protocols can be built into the software, hardware or both.

Without network protocols, computers and other devices would not know how to engage with each other. As a result, except for specialty networks built around a specific architecture, few networks would be able to function, and the internet as we know it wouldn't exist. Virtually all network end users rely on network protocols for connectivity.

How network protocols work: The OSI model
Network protocols break larger processes into discrete, narrowly defined functions and tasks across every level of the network. In the standard model, known as the Open Systems Interconnection (OSI) model, one or more network protocols govern activities at each layer in the telecommunication exchange. Lower layers deal with data transport, while the upper layers in the OSI model deal with software and applications.

To understand how network protocols function, it's crucial to understand the workings of the seven layers of the OSI model:

Physical layer. The physical layer is the initial layer that physically connects two inter-operable systems. It controls simplex or duplex modem transmissions and transfers data in bits. Additionally, it oversees the hardware that connects the network interface card (NIC) to the network, including the wiring, cable terminators, topography and voltage levels.
Data-link layer. The data-link layer is responsible for the error-free delivery of data from one node to another over the physical layer. It's also the firmware layer of the NIC. It puts datagrams together into frames and gives each frame the start and stop flags. Additionally, it fixes issues brought on by broken, misplaced or duplicate frames.
Network layer. The network layer is concerned with information flow regulation, switching and routing between workstations. Additionally, it divides up datagrams from the transport layer into error-free and smaller datagrams.
Transport layer. The transport layer transfers services from the network layer to the application layer and breaks down data into data frames for error checking at the network segment level. This also ensures that a fast host on a network doesn't overtake a slower one. Essentially, the transport layer ensures that the entire message is delivered from beginning to end. It also confirms a successful data transmission and retransmitting of the data if an error is discovered.
Session layer. The session layer establishes a connection between two workstations that need to communicate. In addition to ensuring security, this layer oversees connection establishment, session maintenance and authentication.
Presentation layer. The presentation layer is also known as the translation layer because it retrieves the data from the application layer and formats it for transmission over the network. It addresses the proper representation of data, including the syntax and semantics of information. The presentation layer is also in charge of managing file-level security and transforming data to network standards.
Application layer. The application layer, which is the top layer of the network, oversees relaying user application requests to lower levels. File transfer, email, remote login, data entry and other common applications take place at this layer.
Every packet transmitted and received over a network contains binary data. Most computing protocols add a header at the beginning of each network packet to store information about the sender and the message's intended destination. Some protocols may also include a footer at the end with additional information. Network protocols process these headers and footers as part of the data moving among devices in order to identify messages of their own kind.

When your device attempts to visit a site, there’s a set of rules that it and the servers looking to relay the information back, follow in order to make this transfer of information happen. This is what’s known as the Transmission Control Protocol/Internet Protocol or TCP/IP for short. This set of rules allows for communication between computers to occur. The TCP component to this set of rules involves compiling units of data into what’s known as packets that will be received by another computer. The IP address component focuses on where the packets of data will be sent and received at their respective location. HTTP or Hyper Text Transfer Protocol is the underlying protocol used by the internet and defines how data is formatted and transmitted, and what actions web servers and browsers should take in response to specific commands. But there are a few major concerns, how can we be certain that a malicious third party won’t interfere with this transmission, and second, how can we ensure that all parties involved in the communication are the only ones that should be involved, to begin with?
Happily, there’s a solution to these problems. The first is what’s known as the Secure Sockets Layer or SSL. SSL is a technology that encrypts sensitive data during the transmission thereby making it unreadable to any unauthorized party who may be looking to pry in and steal it. Companies whom you are entrusting your sensitive information, such as credit card numbers or other financial or personal data, purchase what’s known as SSL certificates. These certificates provide authentication for a website and enable the encrypted connection to take place. They also disclose the issuing authority of the certificate as well as the corporate name of the website owner. Transport Layer Security (TSL) certificates provide an even more updated version of the SSL technology. When you see https rather than http appear in your web browser this will indicate that the site is being secured via SSL. The https stands for Hyper Text Transfer Protocol Secure, an extension of the standard transmission HTTP protocol mentioned earlier.
There is another layer of security that exists that’s certainly worth mentioning as it plays a crucial role in this transmission process, firewalls. Firewalls are security devices that monitor incoming and outgoing requests to a network, better known as traffic. Firewalls can either be software or hardware and have the power to decide whether to block or allow certain traffic based on a specific set of rules.
Suppose however a site receives a high volume of requests from users on the internet. Is there a way to help evenly divide these requests between a website’s different servers so one or more of them do not get overwhelmed and possibly crash the site? Thankfully there is such a technology to manage this volume of requests, load balancers. Load balancers, whether in the form of software or hardware, can help evenly distribute the traffic between different servers and provide the flexibility to add or subtract one or more servers as needed. This can allow for a seamless transition if a server is down and needs to be replaced as the user can still access the site since the load balancer can redirect any traffic to only servers that currently can handle them.
Throughout the article, I’ve referred to servers and their role in receiving requests from a user. But what happens when these servers receive these requests and how can they relay the information needed from the site back to the user? The process begins with what’s known as a web server. Assuming that the firewall mentioned earlier has allowed for a legitimate request to the website to be made, the web server then helps process the user’s request by generating a web page for the user to view with the content they requested such as files, videos, or images. Of course, there needs to be a certain “logic” or functionality built into a site that will allow it to deliver this content back to the web server, to begin with. Application servers can handle such requests and run the applications needed to achieve this. The actual data to be sent back such as files or images usually reside and are organized in a database that the application server can access so it can fulfill its role.
It’s hard to imagine that all of the above happens in a matter of seconds to deliver you the content you’re looking for from [www.google.com](http://www.google.com/) or your favorite site. 
This article is written in partial fulfillment of the certificate in software engineer by ALX Africa in collaboration with Halberton schools.

Thanks

souadhelmy:
The first thing that happens when you type [https://www.google.com](https://www.google.com%2A/) in the browser and hit Enter is that the browser checks its DNS (Domain Name System) cache to see if it can find the IP (Internet Protocol) address of the hostname ([www.google.com](http://www.google.com/)). It may also check the operating system DNS cache or any nearby DNS cache because the process has to be really fast. If none of the nearby DNS caches contain the IP address for the hostname, the DNS server at the user’s ISP (Internet Service Provider) would have to perform a recursive DNS lookup to find the IP address. This process involves several recursive queries that would lead to the browser receiving the IP address of the server that would handle the request if the DNS lookup was successful.

PR #1003, opened 2023-07-14
My contribution

Re1gns:
Did a little bit of comprehensive explanation of the TLS handshake

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -369,6 +369,12 @@ This send and receive happens multiple times following the TCP connection flow:
 
 TLS handshake
 -------------
+The TLS (Transport Layer Security) handshake is a crucial process that occurs
+when establishing a secure connection between a client (e.g., web browser) and a server 
+(e.g., website) over the internet. This handshake ensures the confidentiality, integrity,
+and authenticity of the data transmitted between the two entities. 
+Let's take a closer look at the TLS handshake process:
+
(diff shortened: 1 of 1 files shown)


Sampul-CodeMine:
How are HTTP Request handled?

Before a HTTP request is handled and sent to the web server, first of all a resolution of the IP address is gotten from the DNS server which practically is the work of the Authoritative Domain Name Server. When the resolved IP along side a cache data is saved to the browser's cache memory, a request is then created in the form of a datagram containing the request type, the IP address of the client's workstation. Also note that the client's PC connects to the internet using the TCP/IP protocol, the HTTP request is then sent through and the internet to the Load balancer. But this will not fly because google automatically routes any HTTP request "http://www.google.com" through a secured channel to eradicate eavesdropping and secured data. So automatically your HTTP response will be routed through the HTTPS secured protocol and your request will be encrypted using the SSL/TLS encryption protocol.

The load balancer receives the encrypted request and then channel it to a webserver available to handle and process the request. Know that Google's load balancer works also with proximity and location, so a request made in Nigeria for instance will be routed to a google web server in Nigeria. That is why when you make a HTTP request in Nigeria to "http://www.google.com" you could be automatically redirected to "https://www.google.com.ng"

My contribution can also be found here: https://sampulcodemine.hashnode.dev/what-happens-when-you-type-wwwgooglecom-in-your-browser-and-press-enter

PR #1058, opened 2023-07-16
Create 1-what_happen_when_diagram

saria-mohi:


saria-mohi:
create poste that help other student to find simple discuss for what happen when type https://www.google.com

PR #1102, opened 2023-07-17
full description of what happen when you type www.google.com in brows…

Brightsky2394:
Expatiate more on why HTML cannot be parsed using regular top-down or bottom-up parsers, and also on components of browser.


ChinenyeNmoh:
https://medium.com/@chinenyeumeaku/what-happens-when-you-type-https-www-google-com-in-your-browser-and-press-enter-4fb3175790c8  click this link if you want to know what happens when you visit a web page

PR #1138, opened 2023-07-21
Added details on functions of different Nameservers to the DNS Lookup…

jayokeyo:
… section


Aseye-1:
DNS (Domain Name System)

Aseye-1:
DNS servers convert URLs and domain names into IP addresses that computers can understand and use.

PR #1145, opened 2023-07-23
What happens when you type google.com into your browser and press enter?

jayval254:


toamutiltech:
When you type [google.com](http://google.com/) and press enter. What do you think really happened behind the scenes?
I will explain: It happens instantaneously as soon as you hit enter on your browser. The page appears but beneath that, there were a lot of processes that happen from the request to the display of your results.
As soon as you press Enter, your browser initiates what is called a Domain Name System (DNS) request to translate your human-readable domain name "[www.google.com](http://www.google.com/)" into an IP address, which is the actual name and location of the website on the internet. Your browser first checks If there is a cache of that IP address; if it's not found, your browser will send a new request to a public DNS server to get the IP address associated with the domain name. Once the DNS resolution is complete, your browser uses the Transmission Control Protocol (TCP) to establish a connection with the web server hosting Google's website closer to your location because there are severe websites like Google that has more than one server to reduce the loads of request coming from severe devices all over the world, this process is called load balancing. TCP also ensures reliable data transmission by breaking data into packets, numbering them, and reassembling them at the destination. It also handles any data loss or corruption during transit. Also, before your browser establishes a connection with the web server, the data packets must pass through a firewall, which acts as security barrier. Firewalls analyze incoming and outgoing traffic, allowing or blocking packets based on predefined security rules. This helps protect your device and website from any potentially harmful or malicious data. More also when connecting to a secure website like Google, your browser uses HTTPS (Hypertext Transfer Protocol Secure) to establish a secure encrypted connection. This is crucial for protecting sensitive data transmitted between your browser and the web server from eavesdropping or tampering. The SSL (Secure Socket Layer) or its successor, TLS (Transport Layer Security), is responsible for encrypting the data from the server to your devices. When the web server receives your request, it generates an appropriate response. It retrieves the Google homepage and any other requested resources from its file system or cache. The Application server receives a request from the web server then formats the response, also generate any information related to the response from the Database Server and serves it typically in HTML, or any scripting language like PHP, C#, JavaScript, etc, and sends it back to your browser through the established TCP/IP connection. The Application server also generates any information related to the response from the Database Server and serve it to the browser.
When next you type a domain name or [google.com](http://google.com/) on your browser and hit enter, know that severe process took place for the result that you see to appear from the moment you hit enter to the moment a page appears.

PR #1165, opened 2023-09-14
This commit adds a new section to the document explaining the securit…

eddybrownent:
…y measures that browsers take when connecting to websites. It covers HTTPS protocol usage, SSL/TLS certificate validation, mixed content blocking, secure cookies, HSTS, and safe browsing features. These security measures ensure a safe and secure browsing experience for users.


eddybrownent:
mege

PR #1171, opened 2023-09-14
Add explanation for what happens when the 'G' key is pressed futher e…

henry-codex:
…xplaned


henry-codex:
I helped improve the g key section's content i hope that is cool.

PR #1207, opened 2023-09-17
added some info on Parsing HTML, CSS and JS

SAMKE-glitch:
added more detail to the process that occurs behind the scenes of the browser after receiving resources from the server, especially on Parsing HTML, CSS, and JS.


AmrShoukry:
Nice!

PR #1239, opened 2023-09-18
Update README.rst

alantiren:
added more on Window Server


alantiren:
add more on Window Server

PR #1283, opened 2023-10-12
Added an explanantion on browser security

mwendwa-will:
Gave a short description on how the browser handles its security


mwendwa-will:
I've added a new section in the README that covers the essential aspects of browser security.

PR #1305, opened 2023-10-13
One sentence on MAC

Fikayomi31:


Fikayomi31:
Writeup on MAC as a local network

PR #1327, opened 2023-10-13
Update DNS Lookup Section

Y-Baker:


eberechi10:
the DNS lookup and find the Ip address, which is what the computer understands

Y-Baker:
> the DNS lookup and find the Ip address, which is what the computer understands

Right, you simply find the core concept of DNS

PR #1328, opened 2023-10-13
What happens If the DNS server is on a different subnet ?

AmrSamyG247:


Oberon10:
Having a DNS server on a different subnet requires careful network configuration, routing, and security considerations to ensure reliable and efficient DNS resolution for devices across the network.

PR #1389, opened 2023-10-15
Page layout section added

joshua-akuna:


joshua-akuna:
I added the layout section to updated the Readme file

PR #1410, opened 2023-10-15
What happens when you type https://www.google.com on your browser and…

Damicable:
… press enter


Damicable:
I contributed to the work base on the table of contents and more extensively on the work of  Window server

PR #1460, opened 2023-10-16
add more info to the DNS lookup

arkoaikins:


arkoaikins:
More info on the DNS lookup

HenriettaObioha:
What Does DNS Failure Mean?
The DNS failure occurs when users are unable to connect to an IP address via a domain name. A message will pop up that may say “DNS server not available” or “Server DNS Address could not be found.” Essentially, it’s like dialing a phone number and getting back a busy signal. There are multiple potential causes, but the end result is a failed connection.

Potential Sources of a DNS Failure
There are many components that must work together seamlessly to ensure quick web access. DNS failure typically occurs when there is an issue with one of the following:

Router or modem
Browser
Computer
ISP ([Internet Service Provider](https://www.forbes.com/home-improvement/internet/best-internet-providers/))
8 Strategies for Troubleshooting a DNS Failure
1.  Restart Your Software or Device
Sometimes simply exiting the browser completely for a few minutes will solve the problem. You can also try restarting your computer or mobile device. If a plain restart doesn’t work, look up the instructions for restarting in “Safe Mode” to disable extraneous functions that may be impacting your system.

2.  Restart the Modem or Router
Either press the power button or completely unplug the power cord, for at least 20 seconds. Then restart and see if the error is fixed after a few minutes. Also check for any available updates, as these may be needed in order to keep your router running smoothly.

3.  Switch Browsers
If you have been browsing in Google Chrome, try switching over to Safari or Microsoft Edge for a bit. If this doesn’t work you can also try uninstalling, then reinstalling your preferred browser.

4.  Pause Your Firewall
If the DNS error is being caused by your firewall or [antivirus software](https://www.forbes.com/advisor/business/software/best-antivirus-software/), try disabling them temporarily. After turning them off manually, try opening the browser again to see if it’s working. If that did the trick, be sure to reactivate your firewall again right away.

5.  Clear Your Cache
Frequently visited websites are stored, or “cached,” to help them load faster. If you are trying to access a cached site that had a recent IP address update, you may get an error message. Clearing your browser’s cache should correct the problem. Be sure to look up instructions for your specific browser to ensure this is done correctly.

6.  Disable Extra Connections
If there are any extra connections open, such as peer-to-peer features or Bluetooth, disable those and then try restarting the browser.

7.  Keep Everything Updated
Double-check that all software, network adapters and drives are up to date. If any have outstanding updates that need to be completed, run those and then try restarting the browser.

8.  Check DNS Settings
This step will take a bit more effort, as you will need to locate the correct DNS settings via a network administrator or ISP. Once you have the settings, check them against the ones on your computer to confirm they are correct.

If You Can’t Fix a DNS Issue Yourself…
Sometimes it may be necessary to find another resource to help you troubleshoot a DNS error. If you are experiencing frequent problems, the cause could be your internet connection and you will need to contact your internet service provider (ISP). Your provider can check for any service or equipment complications that could be causing issues.

If everything is working correctly on the ISP’s end, it may be time to consider hiring a professional. Both big-box electronic chains and smaller, local IT stores typically offer on-site computer diagnostics and repairs. While this usually means dropping off your PC for a few days, it will be well worth the hassle to have the DNS issue resolved quickly.

PR #1535, opened 2023-12-15
add the TCP/IP to the content

GideonBature:
Add a little note on the TCP/IP (Transmission Control Protocol/Internet Protocol) which is responsible for the whole process, from when a request is made to when a response is received.....


GideonBature:
add TCP/IP to the note

PR #1552, opened 2023-12-16
Created a simple guide for people without technical background

ShegzBit:
This branch contains a .md guide that aims to make the repo available to people with 0 - low technical background featuring a fairytale adventure depicting the process and linking it back to the real process.
This was done to increase the number of people who can benefit from this material, as anyone one with little or no technical background would be overwhelmed by the technical verbosity of the current guide


kegera:
intresting

PR #1573, opened 2023-12-17
Create README.md

rasuliomari:
Adding these details will paint a more complete picture of the intricate journey a DNS request takes across various servers and hierarchical levels before finally reaching the authoritative answer for the desired domain.


rasuliomari:
am ready for any comment in DNS Lookup

PR #1581, opened 2023-12-17
Update README.rst

amakzy:


amakzy:
expanded on the behind the scene of browser

PR #1602, opened 2023-12-17
Update README.rst

Codejuma:


alvoniah:
Great work ssl

Asega-bryan07:
After the connection is ensured to be stable and legal, the browser’s request is sent using an encryption protocol like a Secure Socket Layer (SSL) or Transport Layer Security (TLS) in order to encrypt the data being shared between the device and the server. This is responsible for the “s” in a https request. The s signifies a secure or encrypted.

PR #1652, opened 2024-01-16
Update README.rst

dfunani:
Delve deeper into the role of digital certificates in the TLS handshake.
By expanding on the role of digital certificates in the TLS handshake, we provide a more comprehensive understanding of the security and authentication aspects involved in establishing a secure connection.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -397,6 +397,46 @@ TLS handshake
 * From now on the TLS session transmits the application (HTTP) data encrypted
   with the agreed symmetric key.
 
+Role of Digital Certificates:
+-------------
+In the TLS handshake process, digital certificates play a crucial role in establishing the authenticity of the server and, optionally, the client. Here's an extended explanation:
+
+Server Certificate:
+
+When the client initiates a connection to the server, the server responds by presenting its digital certificate.
+The digital certificate contains the server's public key, information about the server (such as its domain), the digital signature from a Certificate Authority (CA), and the CA's digital signature.
+
+Certificate Authority (CA):
+
+The CA's role is to verify the identity of the server before issuing the digital certificate. CAs are trusted entities that vouch for the authenticity of the information in the certificate.
+The CA's digital signature on the certificate ensures that the certificate has not been tampered with and that the information it contains is valid.
+
+Client Verification (Optional):
+
+While server authentication is mandatory in the TLS handshake, client authentication is optional. If client authentication is requested by the server, the client also presents its digital certificate to the server.
+The client's certificate includes its public key and is also signed by a CA. The server verifies the client's certificate using the CA's public key.
+
+Public Key Exchange:
+
+After the server's identity is verified, the client and server engage in key exchange. The goal is for both parties to agree on a shared symmetric key for encrypting and decrypting data during the session.
+The server's public key, obtained from its certificate, is used by the client to encrypt a pre-master secret and send it securely to the server.
+
+Master Key Derivation:
+
+Both the client and server independently derive the master secret from the pre-master secret and other exchanged parameters. This master secret is used to generate encryption keys for securing the communication.
+
+Forward Secrecy Considerations:
+
+Some modern configurations emphasize the use of forward secrecy. In such cases, the exchanged keying material is used to generate short-lived session keys that are discarded after the session ends, providing an additional layer of security.
+
+Renegotiation and Session Resumption:
+
+TLS supports session resumption for efficiency. If a client and server have previously established a secure session, they can use a session identifier or session ticket to quickly resume that session without repeating the full handshake.
+
+Revocation Checking:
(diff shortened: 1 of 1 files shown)


sallyMohamed:
What happens when you type 'www.google.com' in your browser and press Enter


 You are sitting on the couch sipping your hot cup of coffee in a rather cold weather. And then decided to check out some resources on google. After typing ‘https://www.google.com’ in your browser, you almost immediately see the results of the google home page.

But wait, how did your browser retrieve this page from the google server? A lot of things happen behind the curtain in that innocent request.

In this article, I will walk you through the various processes involved when making requests from your browser and retrieving resources.

You may need to occasionally take a look at the diagram above to better understand the workflow.

 

Domain Name Server (DNS) Request

Whenever you type ‘www.google.com’ – or any other domain name into the search bar, your browser first of all does some IP lookup in its cache. The cache is a small database that stores information about the websites you’ve visited. The browser searches this cache for the IP address of the site you’re currently requesting for. If it finds the IP address there, good! It sends a HTTP request to the server having that IP and the server delivers the content.

But what if your browser can’t find the IP address in its cache? A series of steps are initiated.

·        The browser sends a request to the DNS server to resolve the domain name to an IP address, this is called DNS Resolution.

·        If the DNS resolver does not have a copy of that record, it sends a request to a root nameserver. The root nameservers are servers at the root of the DNS hierarchy.

·        The root nameserver responds with the address of a top-level domain (TLD) nameserver, such a .com or .ng

·        The DNS resolver then sends a request to the TLD nameserver and the TLD nameserver responds with the address of the authoritative nameserver for that domain.

·        The DNS resolver sends a request to the authoritative nameserver and the authoritative nameserver responds with the IP address for the domain.

·        The DNS resolver sends this IP back to the browser.

·        Finally, the browser sends a request at the IP address to retrieve the page.

If the DNS record has been configured to use a load balancer or a Content Delivery Network (CDN), further processes will be involved.

After the lookup, the browser stores the IP records in its cache for easier access next time.

 

TCP/IP Connection

Transmission Control Protocol (TCP) and Internet Protocol are required for a reliable connection between the browser and the server. 

After resolving for the IP address of a server, the web browser sends a connection request to the server. If the server accepts this request, a TCP connection is set. This process is known as a ‘handshake’.

Once the connection is established, the browser can make requests to retrieve pages. The web server then delivers the requested HTML pages. The  browser receives the HTML code and uses them to render the webpage to your screen.

 

Firewall

A firewall is a security system that operates based on some predetermined rules to monitor and control network traffic. It protects against malwares and hackers.

A firewall may be configured based on these two security rules:

·        Rules that allow or block traffic based on the type of traffic.

·        Rules that allow or block traffic based on the source and destination of the request.

Requests can only be accepted by a web server if the firewall rules are met.

 

HTTPS/SSL

HTTPS (HyperText Transfer Protocol Secure) is a secure version of HTTP that is used to send encrypted data between browser and server.

SSL (Secure Socket Layer) and TLS (Transport Layer Security) are protocols used to encrypt data transmission between browser and server.

When a browser makes a request to a website without HTTPS/SSL, all data transmission can be monitored and stollen by a third party and hence security is not guaranteed.

With the HTTPS/SSL server components integrated, all data transmission will be encrypted.

 

Load-balancer

A load balancer is a device that acts as a reverse proxy and distributes network traffic across multiple servers. The major purpose of a load balancer is to reduce traffic load on a server and hence improve the efficiency of the web infrastructure.

When a browser sends a request to a site that uses load balancing, the load balancer distributes the request across a number of servers using a predefined algorithm.

 

Web server

A web server is a computer program that is responsible for handling requests for web pages from clients. The web server returns the appropriate response for the requests that a client makes. 

The web server response typically consists of HTML, CSS and JavaScript files. The browser uses these responses and displays the page to the client.

 

Application server and databases

It is important to note that the web server handles static content but the application server handles dynamic content. When you make a purchase on a website for example, a receipt is generated and displayed to you. That was generated ‘on the fly’ by the application server and sent to the web server. 

The application server may need to make a request to the database in order to generate some pages. When the application generates the necessary page, it sends it back to the web server and the web server forwards the page to the client.

Rendering the Page

A web browser processes the HTML, CSS and JavaScript files in order to render the page. This may include images and appropriate positioning of elements on the page.

There you go. Your request finally gets to your screen.

 

Conclusion

Understanding web infrastructure and the work flow of the internet is necessary as a good software developer.

This blog post is a technical writing task I concluded at ALX Software engineering program.

MohamedAyman128:
When you type "[www.google.com](http://www.google.com/)" in your browser and press Enter, your browser sends a request to the Google servers for the website's content. The Google servers then process the request and send back the website's HTML, CSS, and JavaScript code to your browser, which then renders the website for you to view and interact with.

PR #1745, opened 2024-02-08
Pull request

akmanime:
Hello alx


ahmed-my:
Every domain name like google.com has an equivalent IP address. This IP address can be view using the below syntax
nslookup <domain name>.
for google.com
nslookup google.com will give the IP address 142.250.200.142

PR #1756, opened 2024-02-09
Add compositing step, After painting the render

kavetech2023:


kavetech2023:
This step involves combining multiple layers of the rendered content into a single image that's finally displayed on the screen as pixels.

PR #1823, opened 2024-02-12
I added an article url to explain what happen behind the scenes of Br…

Steveric1:
…owser


Steveric1:
I added a URL to explain what happens behind the scenes when you enter a URL into your browser

PR #1846, opened 2024-03-13
Updated for DevOps

mwanikigachanja:


mwanikigachanja:
Added an option for firefox

PR #1851, opened 2024-03-14
what happens when you type google for dummies

pasebukky:
I created a dummies version of the age old question.


99omniaashraf:
I added a few details to the TLS handshake process, particularly the key exchange processes leading to the generation of the session key.

PR #1881, opened 2024-03-16
adds more detail on "Behind the scenes of the Browser"

rugue:
Added a bit more detail and clarity on "Behind the scenes of the Browser" for ALX Africa Software Engineering Program


Zeezbaba:
https://zeezboss.blogspot.com/2024/03/0x11-what-happens-when-you-type.html

PR #1906, opened 2024-03-17
README.rst

admdmd:


osuxn:
When you type "https://www.google.com" into a web browser's address bar and press Enter, several things typically happen:

DNS Lookup: The browser initiates a Domain Name System (DNS) lookup to translate the human-readable domain name "www.google.com" into an IP address. This process involves querying DNS servers to find the IP address associated with the domain.

Establishing a TCP Connection: Once the IP address is obtained, the browser establishes a Transmission Control Protocol (TCP) connection with the server hosting Google's website. This connection is typically done over port 443 for HTTPS (Hypertext Transfer Protocol Secure) connections.

TLS Handshake (if HTTPS): If the website uses HTTPS, the browser and server engage in a TLS (Transport Layer Security) handshake to establish a secure connection. This involves negotiating encryption parameters, exchanging cryptographic keys, and verifying the server's identity via its SSL/TLS certificate.

HTTP Request: After the connection is established (and secure, if HTTPS), the browser sends an HTTP GET request to the server for the specific resource, in this case, the homepage of Google's website.

Server Processing: The server receives the HTTP request and processes it. In the case of Google's homepage, the server may generate a dynamic HTML page tailored to the user's location, preferences, or other factors.

HTTP Response: The server sends an HTTP response back to the browser, containing the requested resource (in this case, the HTML content of Google's homepage) along with appropriate response headers.

Rendering: The browser receives the HTML content and begins rendering the webpage. It parses the HTML, processes CSS stylesheets, executes JavaScript code (if any), and renders the final webpage layout.

Displaying: Finally, the browser displays the rendered webpage to the user, allowing them to interact with Google's search engine interface and other features.

Francaosawaru:
some of the things that happens when you type "https://www.google.com/" into a web browser's address bar and press enter includes:

1 DNS Lookup
2 Connection
3 HTTP and HTTPS
4 Firewall
5 Load Balancer
6 Webserver
7 Application Server
8 Database
[https://(protocol)](https://jesulayomi.medium.com/(protocol)) www(sub-domain) .google(domain) .com(root) /toys-for-us(path)
Here are the steps in detail…
DNS Lookup
Immediately after hitting enter, the browser sends a request to the domain name system server that returns the IP address of the domain name.
There is a lot more steps in that process that happens in the blink of an eye
The computer/browser checks its A record (A for Address) to find the ip address of [www.google.com.](http://www.google.com./) An A record can be simply described as a table mapping domain names to ip addresses. Websites and webpages are served from a server, these servers are identified by their ip addresses. IP addresses are a unique address that identifies a device on the internet.They are a pain to remember as they are just series of digits or letters (Like 216.58.223.196 — IPv4 or 2001:db8:0:1234:0:567:8:1 - IPv6) so we use domain names instead (Like google.com)
If the IP address is not in the browser’s memory, the request goes through these points.
Browser => Operating system => Resolver => Root => Top Level Domain name servers => IP address
Resolver — Your ISP knows the root (.com, .net, .tech, .org)
Root — Knows the Top level domain name servers ([a-m].gtld-servers.net)
TLD Authoritative name servers — know all domain names that use the root (.com) and returns the IP address of the domain name.
Connection (TCP/IP)
Transmission Control Protocol (TCP) and Internet Protocol (IP) are used to establish a connection between the client and the server hosting the site/application. On some browsers, you’ve probably seen the “Performing TLS handshake” text while a page is loading. The TLS handshake happens when a client sends a request and the server acknowledges the request for a connection After the acknowledgement is sent, the browser then sends the request for the homepage of google.
Then the server sends the html code and any other resources (scripts, images or videos) necessary to display the homepage on the browser over http or https.
Firewall
This monitors outgoing and incoming traffic based on rules (like port restrictions, or specific protocols usage) Some firewalls also check data sent for malwares.A firewall blocks unauthorized traffic and allows only communications that are considered safe, using a set of security rules provided by a network administrator.
Load balancer 
Servers will now be broken down to load balancers and webservers (although some webservers have their own load balancers).
Load balancers are used to share traffic/requests between backend webservers. One popular load balancer software is HA Proxy. It can use different methods for load balancing between webservers such as round-robin, leastconn, ratioed round-robin.
Web server
Webservers serve the static web content to the browsers load is shared between multiple webservers by the load balancer(s).
Application server
An application server is a modern form of platform middleware. This is system software that resides between the operating system (OS), and the external resources (such as a database management system [DBMS], communications, and Internet services). An application server exposes business logic to the clients, which generates dynamic content. It is a software framework that transforms data to provide the specialized functionality offered by a business, service, or application.
Database management system - DBMS
Databases for applications or dynamic content need to be stored, one way of storing is through relational databases that use persistent storage and can perform CRUD operations (create, read, update, delete). The application server requests some data from the google database to use in its response.

PR #1943, opened 2024-03-24
Update README.rst

wang-arshton256:


README.rst
```diff
@@ -666,7 +666,7 @@ Page Rendering
 GPU Rendering
 -------------
 
-* During the rendering process the graphical computing layers can use general
+* During the rendering, process the graphical computing layers can use general
```
the1Riddle, line 669: should the commer be after the process..!

~During the rendering, process the graphical computing layers can use general~
*During the rendering process, the graphical computing layers can use general*

but still, changes can be made based on the main document.

the1Riddle (changes requested):
Could you please recheck it?

wang-arshton256:
Okay, let me do that.

Many thanks.
Wang.

PR #1962, opened 2024-04-10
HTML parsing algorithm

joshkim2441:
Added a high-level overview of the HTML5 parsing algorithm though simple in comparison to the actual, much more detailed algorithm as defined in the HTML5 specification.


joshkim2441:
A simplistic over-view of the `HTML5 parsing algorithm`.

PR #1969, opened 2024-04-11
Additional Details on DNS Resolution

abdelhafid-mahmoudi-env:
One important aspect that is briefly mentioned in the README is DNS resolution. However, it's worth expanding on this topic to provide a clearer understanding of how DNS works in the context of typing a URL into a browser.

When you type "**[https://www.google.com](https://www.google.com)**" into your browser and press Enter, the first step is for the browser to resolve the domain name "www.google.com" to an IP address using the Domain Name System (DNS). This process involves querying DNS servers to obtain the IP address associated with the domain name.

DNS operates hierarchically, with multiple levels of servers involved in the resolution process. The browser typically starts by consulting its own DNS cache to see if it already has the IP address for the domain stored locally. If not, it sends a DNS query to the operating system's configured DNS resolver.

The resolver may have its own cache of DNS records, but if the requested domain's IP address is not found there, it begins a recursive process of querying authoritative DNS servers starting from the root DNS servers, then moving down through the hierarchy of DNS servers responsible for the top-level domain (TLD), the domain itself, and any subdomains.

Each DNS server in the hierarchy responds to the query with either the requested IP address or a referral to another DNS server that may have more specific information about the domain. This process continues until the resolver receives a response with the IP address of the domain.

Once the browser has obtained the IP address of the domain from DNS resolution, it can proceed to establish a connection to the server hosting the website.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -18,6 +18,64 @@ Read this in `简体中文`_ (simplified Chinese), `日本語`_ (Japanese), `한
 (Korean) and `Spanish`_. NOTE: these have not been reviewed by the alex/what-happens-when
 maintainers.
 
+What Happens When...
+====================
+
+This repository is dedicated to unraveling the complexities behind the question: "What happens when you type a URL into your browser's address bar and press enter?" Instead of providing a surface-level answer, we aim to explore each step of the process in detail, leaving no aspect untouched.
+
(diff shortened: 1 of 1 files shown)


abdelaaziz11:
push a pull request

PR #2022, opened 2024-04-14
Add detailed explanation of HTTPS/SSL encryption process

betbaba:
HTTPS/SSL Encryption Process
The HTTPS/SSL encryption process is fundamental in ensuring secure communication between a client (such as a web browser) and a server. It involves several key steps to establish a secure connection:

Initial Request:

When a user types a URL starting with "https://" into their browser and presses Enter, the browser initiates a request to the server using the HTTPS protocol.
SSL Handshake:

The browser sends a request to the server, initiating the SSL handshake process.
During the handshake, the client and server negotiate the parameters of the secure connection, including the SSL/TLS version and the encryption algorithms to be used.
Server Authentication:

The server sends its digital certificate to the client, which contains the server's public key and other identifying information.
The client verifies the authenticity of the certificate to ensure it is issued by a trusted Certificate Authority (CA) and that it is still valid.
Key Exchange:

After authenticating the server, the client generates a random symmetric session key to be used for encrypting and decrypting data during the session.
The client encrypts this session key with the server's public key obtained from the digital certificate and sends it to the server.
Session Encryption:

Upon receiving the encrypted session key, the server decrypts it using its private key.
Both the client and server now share a symmetric session key, which they use to encrypt and decrypt data exchanged during the session.
All subsequent communication between the client and server, including HTTP requests and responses, is encrypted using symmetric encryption.
Secure Data Transfer:

Encrypted data is transmitted between the client and server, protecting it from eavesdropping and tampering by malicious actors.
The symmetric encryption ensures the confidentiality and integrity of the transmitted data.
Session Termination:

Once the session is complete or times out, the symmetric session key is discarded.
Any subsequent sessions between the client and server will require the negotiation of new session keys through a new SSL handshake process.
By following this process, HTTPS/SSL encryption ensures that sensitive information transmitted over the web remains secure and protected from unauthorized access.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -17,7 +17,7 @@ This is all licensed under the terms of the `Creative Commons Zero`_ license.
 Read this in `简体中文`_ (simplified Chinese), `日本語`_ (Japanese), `한국어`_
 (Korean) and `Spanish`_. NOTE: these have not been reviewed by the alex/what-happens-when
 maintainers.
-
+betty
 Table of Contents
 ====================
 


betbaba:
betty add https ssl explanation

PR #2061, opened 2024-04-14
Clarify Factors Influencing Auto-Complete Suggestions

MairaQuillQuest:


MairaQuillQuest:
Clarified that auto-complete suggestions are based on factors such as search history, bookmarks, cookies, and popular internet searches, rather than being solely dependent on the browser's algorithm and private/incognito mode. This change enhances the accuracy of the explanation."

PR #2081, opened 2024-04-15
Contribute to what_happens_when#g_key_is_pressed

emmachinko:
Folks on the Internet have been trying to put together a comprehensive answer to the question on what happens when you type google on browser and press enter

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -622,6 +622,9 @@ any invalid content and go on.
 CSS interpretation
 ------------------
 
+* CSS, or “Cascading Style Sheets,” is used for styling and laying out webpages. 
+  It can be used to adjust content size, spacing, color and font or add decorative features, 
+  such as animations or split content into columns.
 * Parse CSS files, ``<style>`` tag contents, and ``style`` attribute
   values using `"CSS lexical and syntax grammar"`_
 * Each CSS file is parsed into a ``StyleSheet object``, where each object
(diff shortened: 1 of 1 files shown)


SirJupiter:
1. Physical Keystroke:

You press down on the physical "g" key on your keyboard. This action depresses the keycap and registers a signal with the keyboard's internal circuitry.

2. Scan Code Generation:

The keyboard translates the physical key press into a unique scan code. This code identifies the specific key that was pressed. Different keyboards may use slightly different scan code schemes, but the essential purpose remains the same.

3. Interrupt Signal to CPU:

The keyboard controller sends an interrupt signal to the CPU, indicating that a key has been pressed and requires attention. This temporarily halts the CPU's current task and prompts it to handle the keyboard input.

4. Interrupt Handler Intervention:

The operating system (OS) has a dedicated program called an interrupt handler that specifically deals with input from the keyboard. This handler receives the scan code from the keyboard controller.

5. Scan Code to Key Mapping:

The interrupt handler utilizes a lookup table or similar mechanism to translate the received scan code into a meaningful key identifier. In this case, the scan code for "g" is mapped to the character "g".

In essence, pressing the "g" key initiates a sequence of events involving the keyboard hardware, the operating system's interrupt handling mechanism, and the application in focus. The specific outcome hinges on the context (application, modifier keys used) and how the software interprets the combination of key presses.

PR #2110, opened 2024-04-23
Update README.rst

Akanni23:
(https://github.com/alex/what-happens-when/files/15078657/What.happens.when.you.type.an.URL.in.the.browser.and.press.Enter.docx)


Akanni23:
what happen when you type . . .

PR #2143, opened 2024-05-10
what happens when you press g

Golhun:
Updated the section on what happens when you press "g".


Golhun:
Updated he section on what happens when you press g

PR #2172, opened 2024-05-12
what happens

Tiedaman33:
additional explation on the G-keys


GithinjiJustus:
Explaining the G- keys

PR #2210, opened 2024-05-12
Add description of what happens if the domain name is an alias

Zennoon:
This pull request is an attempt to include a description of what would happen if the domain name in the DNS query is an alias to another domain name, in which case a new DNS query is initiated.


firacil:
If the domain name in the DNS query is an alias (known as a CNAME record) to another domain name, the DNS resolver will receive the alias (CNAME) record in the DNS response. Subsequently, it will initiate a new DNS query to resolve the canonical (actual) domain name associated with the alias.

Zennoon:
I agree, That is what I put in the commit (the change I made).

PR #2228, opened 2024-05-12
Update README.rst

MedBens02:
Continuing from our exploration of what happens when you type "google.com" and press enter, let's delve into a modern web capability that further enhances user experience: the Network Information API.


mihaimaruseac (changes requested):
This should not be at the end of the file, it breaks links.

PR #2255, opened 2024-05-13
Added the most used encryption algorithm

NwaforChukwuebuka:
RSA algorithm has been tested and proven by the NSA to be unbreakable.


NwaforChukwuebuka:
Added RSA as a means of  symmetric encryption used for secure transfer of data

PR #2261, opened 2024-05-16
Update README.rst

Iganza-roy:
Refined DNS lookup explanation


Iganza-roy:
Updated the explanation regarding DNS lookup to provide clarity on the ARP process involved when the DNS server is on a different subnet. The commit enhances the understanding of how the network library obtains the MAC address of the DNS server or default gateway IP, facilitating communication at the data link layer.

PR #2285, opened 2024-05-25
additional info

VictorNalu:


VictorNalu:
I added some new information; Firewall and Load Balancer. Please check it. Thank you.

PR #2312, opened 2024-06-19
Create how the browser handle HTTP and HTTPS

Zeugen23:
In this modifications i explain on how the browser way of Handling HTTP differs from HTTPS, underlying security mechanisms and the way the data is transmitted.


softontechsolution:
Wrote and published an article explaining how the client request is sent to the local DNS, from there to the DNS resolver, ETC to get the server IP address thereafter sending the server IP to the client.
I also considered a situation where a firewall is installed or some form of load balancing is on the server.

PR #2353, opened 2024-07-19
How the serial signals/data packets are decoded and interpreted

Yermehyaw:
A step wise description of how the serial signals received by the host PC is converted/decided into data packets by the host USB controller. 
If the added style of editing isn't preferable, it can be easily adjusted


Yermehyaw:
Also now includes how these data packets are then interpreted by the HID universal keyboard device driver into responses by the OS

JoshuaAisien (approved):

PR #2358, opened 2024-07-20
Added explanation for TCP congestion control

Dev-Kings:


Dev-Kings:
Added more explanation on what happens when packets are dropped.

PR #2394, opened 2024-07-21
How DNS works

Bo1atit0:
Added a  "How DNS works" section which simplifies the explanation of how DNS works with a simple analogy


iFingers007:
Is there no way a picture can be added for a better illustration?

PR #2447, opened 2024-07-26
DNS Resolution

patrickodeh1:
This pull request adds a brief section on DNS resolution to the "what-happens-when" repository. The new content provides an in-depth explanation of how the domain name system translates human-readable domain names into IP addresses, covering key concepts such as:

Recursive vs. Iterative Queries: Explanation of how DNS queries are processed, with a distinction between recursive queries initiated by the client's DNS resolver and iterative queries performed by DNS servers.

Caching Mechanisms: Details on how DNS responses are cached at multiple levels (browser, operating system, ISP) to improve lookup efficiency and reduce latency for repeated requests.

DNS Hierarchy: Overview of the DNS hierarchy, including root servers, top-level domain (TLD) servers, and authoritative servers, illustrating the path a DNS query takes to resolve a domain name.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -367,6 +367,23 @@ This send and receive happens multiple times following the TCP connection flow:
    * The other sides ACKs the FIN packet and sends its own FIN
    * The closer acknowledges the other side's FIN with an ACK
 
+DNS Resolution
+--------------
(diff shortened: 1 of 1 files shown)


patrickodeh1:
DNS Resolution

PR #2455, opened 2024-07-27
HTTP Server Request Handle update

carrier-pigeon-san:
* Generation of HTTP response headers by server.
* Logging of requests and responses for monitoring and debugging.


carrier-pigeon-san:
This pull request serves as an answer to a task for an ALX Software Engineering project.

PR #2461, opened 2024-07-28
Update README with details on HTTP protocol versions

Kadesa-Star:
This Updates README with details on HTTP protocol versions


Kadesa-Star:
Great

PR #2474, opened 2024-07-29
what happens when

hyiaman:
I added a blog and brief technical description of what happens when from my own perspective.


hyiaman:
Kindly check out and approve

tobidelly:
Checked

PR #2479, opened 2024-08-23
Added Some Notes 

Ziad-Khaled11:
Added some more detailed information in the part of Behind the scenes of the Browser
Hope this makes help in understanding it more


Ziad-Khaled11:
Done

